PR #28, opened 2022-02-06
Use a more portable method of detecting the current script's directory

jthurne:
THe method used before relied on GNU readlink's -e CLI flag, which isn't
supported on the version of readlink shipped with BSD and MacOs.

Finding the current script's directory is tricky business. There are many
StackOverflow questioins on this very topic, and many opinions on the "best"
way to do it. The solution in this commit was carefully selected after
reviewing many sources of information on the topic. Below are references to the
most useful sources:

	* [Reliable way for a Bash script to get the full path to itself [duplicate]](https://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself):
    This is the best question on Stack Overflow because the question contains a
    specific set of requiremewnts (which are very close to our own requirements).
    In particular, [this answer](https://stackoverflow.com/a/9107028/106189) is helpful.

  * [Determining the Location of Your Script in Bash](https://www.binaryphile.com/bash/2020/01/12/determining-the-location-of-your-script-in-bash.html):
    This is an excellent blog post which walks through and explains the solution in detail.


etiennestuder:
Thanks, @jthurne .

etiennestuder (approved):

etiennestuder:
@jthurne I cannot review it technically but I'm sure this is an improvement if you raise it. Feel free to merge this.

Just curious: was this PR motivated by someone reporting an issue?

davidmc24 (approved):

PR #41, opened 2022-03-02
Add build scan publication.

runningcode:
Note: This does not add the access token to the `production-release.yml` workflow.


etiennestuder (approved):

.github/workflows/development-release.yml
```diff
@@ -10,6 +10,8 @@ jobs:
   development_release:
     name: Release
     runs-on: ubuntu-latest
+    env:
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
```
etiennestuder, line 14: We need to do the same for the production-release.yaml

PR #45, opened 2022-03-11
Add composite actions to run the experiments

jprinet:
Relates to https://github.com/gradle/ge/issues/13584


.github/composite/exp2/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    type: string
+
+runs:
+  using: "composite"
+  steps:
+    - name: Run Gradle Experiment 2
+      run: |
+        cd gradle-enterprise-gradle-build-validation
+        ./02-validate-local-build-caching-same-location.sh -r ${{ inputs.repositoryUrl }} -b ${{ inputs.branch }} -t ${{ inputs.task }} -s ${{ inputs.gradleEnterpriseUrl }}
+      shell: bash
```
jthurne, line 24: It might be useful to make the shell configurable. For example, this would allow us to use these actions in our cross-platform tests (where we sometimes have to use a shell that is different from bash).

jthurne (commented):
These are super-cool!

.github/composite/exp2/action.yml
jprinet: I would have been happy to do so but the templated syntax does not apply for `shell`.

What is the exact use case @jthurne as the so called `bash` _shell_ supports all platforms (_Windows_ included):
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

> The default shell on non-Windows platforms with a fallback to sh. When specifying a bash shell on Windows, the bash shell included with Git for Windows is used.

jprinet:
> I think we want these to exist in this repository: https://github.com/gradle/gradle-enterprise-build-optimization-experiments
My impression is that the repo providing the scripts should be the one providing the composite actions to use them. This would reduce the risk of desynchronization.

> The other requirement (from the ticket) is that we want each project to be its own Github workflow. This will allow better re-use of these from different teams so they can copy them to their use case. That being said, this might also be suitable. I will let @etiennestuder decide.
So far, there is no need to separate the workflows per project, _Androidx_ might become special though. Having one single skeleton to maintain would reduce code duplication.

.github/composite/exp2/action.yml
jthurne: > I would have been happy to do so but the templated syntax does not apply for `shell`.

Oh yeah, that's right. I ran into that with the cross-version tests as well.

> What is the exact use case @jthurne as the so called bash shell supports all platforms (Windows included):

The shell setting of "bash" on Windows uses the version of bash that comes with Git on Windows, but it isn't a complete bash implementation. Most people who use Windows instead use a full Linux version of bash under the Windows Linux Subsystem (WSL).  It isn't possible to redefine the `bash` in the GitHub Actions to refer to the WSL bash instead. So instead, we need to use `wsl-bash` for our Windows tests (see [the Setup WSL GH Action](https://github.com/marketplace/actions/setup-wsl)).

Since we cannot make it configurable, it may be better not to set a `shell` at all. On most platforms, `bash` is the default shell anyway. On other platforms, like Windows, not setting a shell would allow the user to set the default shell to something else (like [what is done in our cross platform tests](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/.github/workflows/cross-platform-testing.yml#L32)).

jprinet: The `shell` instruction is mandatory in a composite action, this being said, this should be ok to have `bash` for the steps executed within the composite action and another `shell` for other steps. wdyt?

jthurne: > The shell instruction is mandatory in a composite action

That's a bummer.

> this should be ok to have bash for the steps executed within the composite action and another shell for other steps

For our cross platform tests, it's important that the actual validation scripts are run in the WSL bash rather than Git Bash. If we cannot rely on the default shell configured for the workflow, then that just means we won't be able to use these composite actions with our cross platform tests. 

But that's just one GitHub workflow. It still means they can be used across the open source projects, so there's still value.

jprinet: Yep, just updated my comment meanwhile...
I think we could try to find a way with some conditional steps, let me come with a proposal.

etiennestuder:
@jprinet let's please start simple. there is no need to do more than what we want for the OSS projects. it's also much simpler to review and evolve this way.

jprinet:
My idea was to have an optional input when requiring `wsl-bash` however, the step just fails when the interpreter is not present.
We could have some dedicated composite with the `wsl-bash`, which would require the caller to handle the conditional call.

> @jprinet let's please start simple. there is no need to do more than what we want for the OSS projects. it's also much simpler to review and evolve this way.

Absolutely @etiennestuder, we'll keep it as is for the moment and we can consider integrating with the cross platform tests as another task.

.github/composite/exp1/action.yml
```diff
@@ -0,0 +1,49 @@
...
+      id: run
+      run: |
+        cd gradle-enterprise-gradle-build-validation
+
+        # add instructions to save build scan urls
+        BUILD_SCAN_FILE_1=$(pwd)/exp1-build-scan-url-1.txt
+        BUILD_SCAN_FILE_2=$(pwd)/exp1-build-scan-url-2.txt
+        echo "echo \${build_scan_urls[0]} > ${BUILD_SCAN_FILE_1}" >> 01-validate-incremental-building.sh
+        echo "echo \${build_scan_urls[1]} > ${BUILD_SCAN_FILE_2}" >> 01-validate-incremental-building.sh
+
```
etiennestuder, line 42: This can be simplified since running the experiment will create files that contain the build scan ids already.

etiennestuder: We should consider adding the entire experiment receipt created by the validation scripts as an asset/resource attached to the GH workflow execution.

etiennestuder (changes requested):

.github/composite/getLatest/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Get latest release
```
etiennestuder, line 1: I would rename the folder in which this file lives (and thus the name of the action) to `download`.

.github/composite/README.md
```diff
@@ -0,0 +1,38 @@
+# Composite Github Actions
```
etiennestuder, line 1: I would change the folder hierarchy as following:
```
.github/
    actions/
        gradle/
            exp1/
            exp2/
            exp3/
            download/
        maven/
            exp1/
            exp2/
            exp3/
            download/
```

.github/composite/getLatest/action.yml
etiennestuder: I'll resolve this comment since it is already covered by my comment on the README file where I propose a different folder herarchy.

.github/composite/exp1/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    description: "Project directory"
+    required: false
+    default: .
+  gradleEnterpriseUrl:
+    description: "Gradle Enterprise URL"
+    required: true
+outputs:
+  buildScanId1:
+    description: "First build scan identifier"
+    value: ${{ steps.run.outputs.buildScanId1 }}
```
etiennestuder, line 24: Is it just the build scan id or the entire build scan URL? (the latter would make more sense to me) If so, I would rename the field to buildScaFirstBuild and buildScanSecondBuild to be aligned with the naming we use in the output of the underlying validation scripts.

etiennestuder (changes requested):

.github/composite/README.md
```diff
@@ -0,0 +1,38 @@
...
+Create a Github Action workflow, fulfilling the build requirements (add JDK...) and then add the following steps (replacing the placeholders):
+
+```yaml
+steps:
+  # Download scripts latest version
+  - uses: gradle/gradle-enterprise-build-validation-scripts/.github/composite/getLatest@v1.0.2
+    with:
+      token: ${{ secrets.GITHUB_TOKEN }}
+  # Run experiment 1
+  - uses: gradle/gradle-enterprise-build-validation-scripts/.github/composite/exp1@v1.0.2
```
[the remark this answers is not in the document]
jprinet: This is a tag to be created, as composite actions are retrieved by repo tag. We will adjust it before merging.

.github/composite/README.md
jprinet: done, please note that Maven is not yet in the scope of this PR

.github/composite/exp1/action.yml
jprinet: This is the build scan identifier as it used down the line to query the REST API

.github/composite/exp1/action.yml
jprinet: > This can be simplified since running the experiment will create files that contain the build scan ids already.

This is not so easy to retrieve the receipt as it is part of a randomly named generated folder.
In addition to that, even if unlikely, you could think of a case where several runs could happen for an experiment, forcing to sort the generated folders by timestamp.
Having a way to set as an argument the absolute path to the receipt could be another way to get the job done.

> We should consider adding the entire experiment receipt created by the validation scripts as an asset/resource attached to the GH workflow execution.

The only thing we need from the receipt is actually the build scan id to query the REST API later on

.github/composite/exp1/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  buildScanId2:
+    description: "Second build scan identifier"
+    value: ${{ steps.run.outputs.buildScanId2 }}
+
+runs:
+  using: "composite"
```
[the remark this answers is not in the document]
jprinet: This is to define a reusable workflow piece (composite)
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

.github/composite/exp1/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  steps:
+    - name: Run Gradle Experiment 1
+      id: run
+      run: |
+        cd gradle-enterprise-gradle-build-validation
```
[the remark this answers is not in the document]
jprinet: indeed this is the default folder name in which the archive is extracted

.github/composite/exp1/action.yml
jthurne: It's possible to get the experiment directory from the summary printed at the end of the experiment. In that directory is a the receipt, but there is also a file that contains just the build scan URLs.

.github/composite/exp1/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  task:
+    description: "Gradle task"
+    required: true
+  projectDir:
+    description: "Project directory"
+    required: false
+    default: .
+  gradleEnterpriseUrl:
+    description: "Gradle Enterprise URL"
+    required: true
```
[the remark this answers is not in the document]
jprinet: done, kept `gitRepo` and `tasks` as mandatory though

.github/composite/exp1/action.yml
jprinet: I'd like not to parse the summary, looking for Build scan first build / Build scan second build sounds fragile. 
Looking into the `csv` might lead to some unexpected refactoring implication at some point imo.

Having a way to set the experiment folder sounds more solid from my perspective.

.github/composite/getLatest/action.yml
```diff
@@ -0,0 +1,17 @@
...
+description: "Install Scripts latest release"
+
+inputs:
+  token:
+    description: "Github token"
```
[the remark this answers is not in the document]
jprinet: I made it optional

.github/composite/exp1/action.yml
etiennestuder: Now that we have the symlink to the latest experiment, we can get the scan URLs from the csv. (once we release 1.0.2)

PR #48, opened 2022-03-18
Remove unnecessary pushd/popd

jprinet:
When using `project-dir` parameter, build fails as relative path to `init-scripts` is incorrect.
This is happening because we `pushd` twice to the `project-dir` folder
https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/bd457beeba9c57d8e8e2afcf396693e2f3be104d/components/scripts/lib/gradle.sh#L13
and then
https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/bd457beeba9c57d8e8e2afcf396693e2f3be104d/components/scripts/lib/paths.sh#L51

Signed-off-by: Jerome Prinet <jprinet@gradle.com>

### DIFF
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,1 @@
-
+* When using the ```project-dir``` argument to specify the build invocation directory, script execution was failing due to wrong ```pushd``` ordering in ```invoke_gradle```
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,1 @@
-* When using the ```project-dir``` argument to specify the build invocation directory, script execution was failing due to wrong ```pushd``` ordering in ```invoke_gradle```
+* FIX: Gradle experiment scripts fail when using the ```--project-dir``` argument to specify the build invocation directory.
--- a/components/scripts/lib/gradle.sh
+++ b/components/scripts/lib/gradle.sh
@@ -10,10 +10,10 @@ invoke_gradle() {
   local args
   args=()
 
-  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." 3
-
   local init_scripts_dir
   init_scripts_dir="$(init_scripts_path)"
+
+  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." 3
 
   if [ "$enable_ge" == "on" ]; then
     args+=(--init-script "${init_scripts_dir}/enable-gradle-enterprise.gradle")


components/scripts/lib/paths.sh
```diff
@@ -47,15 +47,9 @@ relative_path() {
 }
 
 relative_lib_path() {
-  #shellcheck disable=SC2164  # We will handle the error when we try to invoke the build
-  pushd "${project_dir}" > /dev/null 2>&1
```
jthurne, line 51: Unfortunately, I believe this change would break experiments 2, 3, and 5 in some cases because they also call this function, but outside of the context of the subproject.

More specifically, experiments 2, 3, and 5 call the `init_scripts_path` function, which calls this function. See [this in experiment 2](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/components/scripts/gradle/02-validate-local-build-caching-same-location.sh#L116) for an example.

I believe a more reliable fix would be to [move the call to init_script_path in gradle.sh](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/components/scripts/lib/gradle.sh#L15-L16) to be [before it pushds to the project dir](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/components/scripts/lib/gradle.sh#L13) (move lines 15 and 16 to be before line 13).

That would be a quick-fix anyway. Another option is to update the experiments to not build the relative path to the additional init scripts in the experiment scripts and instead pass in the names of "extra" init scripts into the `invoke_gradle` function and calculate their relative paths there instead.  

Either fix would be fine with me.

jthurne (changes requested):

jthurne (approved):

jthurne:
@jprinet I talked to Etienne and we're fine if you want to merge this now. Please also add a note to [release/changes.md](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/release/changes.md) so that we remember this fix when we release the next version of the scripts.

jprinet:
> @jprinet I talked to Etienne and we're fine if you want to merge this now. Please also add a note to [release/changes.md](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/release/changes.md) so that we remember this fix when we release the next version of the scripts.

done! Please let me know if you're fine with the wording.

release/changes.md
```diff
@@ -1 +1 @@
-
+* Fix ```pushd``` ordering in ```invoke_gradle```
```
jthurne, line 1: This will end up in the release notes for the release, so we should make sure this is something someone using the scripts would understand. As such, I suggest rewording this to describe the bug that this change fixes (something about how specifying a project directory was broken when running the Gradle experiments).

jprinet: Sure thing, I updated the comment!

release/changes.md
```diff
@@ -1 +1 @@
-
+* When using the ```project-dir``` argument to specify the build invocation directory, script execution was failing due to wrong ```pushd``` ordering in ```invoke_gradle```
```
jthurne, line 1: This is better, but it still leaks implementation details and doesn't really tell the user what was broken. Maybe something like this?

```suggestion
* FIX: Gradle experiment scripts fail when using the ```--project-dir``` argument to specify the build invocation directory.
```

jthurne, line 1: This is better, but it still leaks implementation details and doesn't really tell the user what was broken. Maybe something like this?

```suggestion
* FIX: Gradle experiment scripts fail when using the ```--project-dir``` argument to specify the build invocation directory.
```

PR #63, opened 2022-04-05
Composite action to assert caching

jprinet:
A composite step which can be reused in any Github workflow to check the outcome of the experiments triggered by the automated scripts.
The step checks that both builds of an experiment succeeded and that the second one is matching a number of expected tasks to run (check against GE API).

Signed-off-by: Jerome Prinet <jprinet@gradle.com>


etiennestuder:
@jprinet I'll close this PR given we will rely on the scripts themselves to verify if all cacheable tasks are taken from cache or not.

PR #65, opened 2022-04-08
Enable configuration caching

clayburn:
This change requires removing 2 invocations of `Task.project` at
execution time used to access the project version for the
`copyGradleScripts` and `copyMavenScripts` tasks

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -93,13 +93,16 @@ val copyGradleScripts = tasks.register<Copy>("copyGradleScripts") {
     group = "build"
     description = "Copies the Gradle source and generated scripts to output directory."
 
+    inputs.property("project.version", project.version)
+    val projectVersion = project.version.toString()
+
     from(layout.projectDirectory.file("LICENSE"))
     from(layout.projectDirectory.dir("release").file("version.txt"))
     rename("version.txt", "VERSION")
 
     from(layout.projectDirectory.dir("components/scripts/gradle")) {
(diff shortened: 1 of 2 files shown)


etiennestuder (approved):

build.gradle.kts
```diff
@@ -93,13 +93,16 @@ val copyGradleScripts = tasks.register<Copy>("copyGradleScripts") {
     group = "build"
     description = "Copies the Gradle source and generated scripts to output directory."
 
+    inputs.property("project.version", project.version)
```
etiennestuder, line 96: @clayburn Can you explain the motivation to declare the project version a task inputs property? While correct, it seems related to build caching and not to configuration caching.

clayburn: Maybe it isn't strictly needed for the configuration cache, but we do mention this as a strategy in the configuration cache document I was following:

> For example, if you reference a Project in order to get the project.version at execution time, you should instead directly declare the project version as an input to your task using a Property<String>.

[Configuration Cache - Certain types must not be referenced by tasks](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:disallowed_types)

etiennestuder: @runningcode what do you mean by `you could saw lines 96 and 97`? Did you mean `switch the lines`?

runningcode: Yes, correct.

PR #66, opened 2022-04-11
Add java toolchain support to buildSrc

alextu:
- Adding toolchain to buildSrc
- Changing existing jdk toolchain version from 8 to 11, to match Github action's provisionned version

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,5 +1,11 @@
 plugins {
-    kotlin("jvm") version "1.5.21"
+    kotlin("jvm") version "1.6.0"
+}
+
+kotlin {
+    jvmToolchain {
+        (this as JavaToolchainSpec).languageVersion.set(JavaLanguageVersion.of(8))
+    }
 }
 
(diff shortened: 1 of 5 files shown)


buildSrc/build.gradle.kts
```diff
@@ -1,5 +1,11 @@
 plugins {
-    kotlin("jvm") version "1.5.21"
+    kotlin("jvm") version "1.6.0"
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(11))
+    }
 }
```
etiennestuder, line 9: Please use Java 8 in all 3 occurrences. We want the scripts to be Java 8 comaptible, just like the reast of BT and GE.

etiennestuder (changes requested):

buildSrc/build.gradle.kts
alextu: Sounds good, I've updated the Github action workflows files accordingly

etiennestuder (approved):

PR #70, opened 2022-04-12
Use existing appVersion variable

runningcode:


clayburn (approved):
Yeah that looks better, approved.

PR #73, opened 2022-04-18
Enable configuration caching

clayburn:


build.gradle.kts
```diff
@@ -126,20 +127,21 @@ val copyMavenScripts = tasks.register<Copy>("copyMavenScripts") {
     group = "build"
     description = "Copies the Maven source and generated scripts to output directory."
 
-    inputs.property("project.version", appVersion)
+    val projectVersion = appVersion
```
clayburn, line 130: Note: had to copy this into a local variable in order to maintain configuration caching.

build.gradle.kts
```diff
@@ -93,6 +93,7 @@ val copyGradleScripts = tasks.register<Copy>("copyGradleScripts") {
     group = "build"
     description = "Copies the Gradle source and generated scripts to output directory."
 
+    val projectVersion = appVersion
     inputs.property("project.version", appVersion)
```
etiennestuder, line 97: I would use `projectVersion` here instead of `appVersion`.

build.gradle.kts
```diff
@@ -260,6 +262,11 @@ tasks.register<CreateGitTag>("createReleaseTag") {
 
 tasks.named("githubRelease") {
     dependsOn("createReleaseTag")
+    notCompatibleWithConfigurationCache("$name task does not support configuration caching")
```
etiennestuder, line 265: Same comment as on the CCUD plugin.

etiennestuder (changes requested):

build.gradle.kts
etiennestuder: I suggest to add this as a comment, e.g.

```
    val projectVersion = appVersion. // local variable required for configuration cache compatibility
```

Please do the same for Maven further down, too.

PR #81, opened 2022-05-02
Fix tasks argument containing whitespaces

jprinet:
Tasks argument has to be protected to allow whitespaces

### DIFF
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -42,31 +42,36 @@ runs:
       run: |
         cd gradle-enterprise-gradle-build-validation
 
-        EXTRA_ARGS=""
+        # Need to process separately tasks argument as it may contain whitespaces
+        ARG_TASKS=""
+        if [ ! -z "${{ inputs.tasks }}" ]; then
+          ARG_TASKS="-t ${{inputs.tasks }}"
+        fi
+
+        # Unprotected arguments
+        UNPROTECTED_ARGS="-r ${{ inputs.gitRepo }}"
         if [ ! -z "${{ inputs.gitBranch }}" ]; then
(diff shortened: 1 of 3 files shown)


.github/actions/gradle/experiment-1/action.yml
```diff
@@ -42,31 +42,36 @@ runs:
...
+          UNPROTECTED_ARGS="$UNPROTECTED_ARGS -s ${{ inputs.gradleEnterpriseUrl }}"
         fi
         if [ ! -z "${{ inputs.enableGradleEnterprise }}" ]; then
-          EXTRA_ARGS="$EXTRA_ARGS -e"
+          UNPROTECTED_ARGS="$UNPROTECTED_ARGS -e"
         fi
 
         # run experiment
-        ./01-validate-incremental-building.sh -r ${{ inputs.gitRepo }} $EXTRA_ARGS
+        ./01-validate-incremental-building.sh "${ARG_TASKS}" ${UNPROTECTED_ARGS}
```
etiennestuder, line 74: Directories can also contain whitespace.

etiennestuder: Isn't there a way to "protect" the arguments of each option separately? It does not seem "right" that we do this special casing just for tasks, and that it percolates all the way to cmd invocation here.

etiennestuder: I see. It might be better to add each option separately, always protected by quotes, just like you did for tasks, but also for all other options.

jthurne: > It might be better to add each option separately, always protected by quotes

Yeah, I agree. That would also make it much easier to see how each argument is quoted, which is difficult to see and understand in the current implementation.

PR #96, opened 2022-06-08
Switch to the OpenAPI Generator 6.0.0 Gradle Plugin

jthurne:
This replaces the old Swagger Generator Gradle plugin. It has the following
benefits:

  * More closely follows normal Gradle plugin conventions (for example, there isn't the weird swaggerSources container anymore, and there are instead more normal convention objects for the tasks the plugin provides).

  * Is fully incremental building, build caching, and configuration-cache compatible.

This commit also switches to using an .openapi-generator-ignore file to exclude
the files we don't need to have generated. The file has to be at the root of
the module otherwise it does not appear to work correctly, and the entries
within the file must be relative to the project root.


sorin-florea (approved):

components/fetch-build-scan-data-cmdline-tool/build.gradle.kts
```diff
@@ -35,28 +33,31 @@ java {
...
+        "useBeanValidation" to "false",
+        "disallowAdditionalPropertiesIfNotPresent" to "false",
+        "sourceFolder" to ""  // makes IDEs like IntelliJ more reliably interpret the class packages.
+    ))
+}
+
+sourceSets {
+    main {
+        java {
+            srcDir(tasks.findByName("openApiGenerate")!!)
```
etiennestuder, line 60: findByName resolves the task early:
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview

components/fetch-build-scan-data-cmdline-tool/build.gradle.kts
```diff
@@ -1,26 +1,24 @@
-import org.hidetake.gradle.swagger.generator.GenerateSwaggerCode
```
etiennestuder, line 1: Global comment: please verify that the build is still configuration cache compatible.

etiennestuder (changes requested):

components/fetch-build-scan-data-cmdline-tool/build.gradle.kts
jthurne: ✅ Good catch!  Fixed.

components/fetch-build-scan-data-cmdline-tool/build.gradle.kts
jthurne: ✅ Verified. The build is entirely configuration cache compatible.

build.gradle.kts
```diff
@@ -180,12 +180,12 @@ val assembleMavenScripts = tasks.register<Zip>("assembleMavenScripts") {
...
-tasks.named("assemble") {
-    dependsOn("assembleGradleScripts")
-    dependsOn("assembleMavenScripts")
+tasks.assemble {
+    dependsOn(assembleGradleScripts)
+    dependsOn(assembleMavenScripts)
 }
 
-tasks.register<Shellcheck>("shellcheckGradleScripts") {
+val shellcheckGradleScripts = tasks.register<Shellcheck>("shellcheckGradleScripts") {
```
britter, line 188: You can write this like this, so you don't need to repeat the name twice:

```suggestion
val shellcheckGradleScripts by tasks.registering(Shellcheck::class) {
```

build.gradle.kts
```diff
@@ -206,7 +206,7 @@ tasks.register<Shellcheck>("shellcheckGradleScripts") {
     }
 }
 
-tasks.register<Shellcheck>("shellcheckMavenScripts") {
+val shellcheckMavenScripts = tasks.register<Shellcheck>("shellcheckMavenScripts") {
```
britter, line 209:
```suggestion
val shellcheckMavenScripts by tasks.registering(Shellcheck::class) {
```

build.gradle.kts
```diff
@@ -227,9 +227,9 @@ tasks.register<Shellcheck>("shellcheckMavenScripts") {
     }
 }
 
-tasks.named("check") {
-    dependsOn("shellcheckGradleScripts")
-    dependsOn("shellcheckMavenScripts")
+tasks.check {
+    dependsOn(shellcheckGradleScripts)
+    dependsOn(shellcheckMavenScripts)
```
britter, line 232:
```suggestion
    dependsOn(shellcheckGradleScripts, shellcheckMavenScripts)
```

build.gradle.kts
```diff
@@ -180,12 +180,12 @@ val assembleMavenScripts = tasks.register<Zip>("assembleMavenScripts") {
     into(archiveBaseName.get())
 }
 
-tasks.named("assemble") {
-    dependsOn("assembleGradleScripts")
-    dependsOn("assembleMavenScripts")
+tasks.assemble {
+    dependsOn(assembleGradleScripts)
+    dependsOn(assembleMavenScripts)
```
britter, line 185:
```suggestion
    dependsOn(assembleGradleScripts, assembleMavenScripts)
```

build.gradle.kts
```diff
@@ -255,13 +255,13 @@ githubRelease {
     releaseAssets(assembleGradleScripts, assembleMavenScripts, generateChecksums.get().outputs.files.asFileTree)
 }
 
-tasks.register<CreateGitTag>("createReleaseTag") {
+val createReleaseTag = tasks.register<CreateGitTag>("createReleaseTag") {
```
britter, line 258:
```suggestion
val createReleaseTag by tasks.registering(CreateGitTag::class) {
```

components/capture-build-scan-url-maven-extension/build.gradle.kts
```diff
@@ -20,6 +20,6 @@ java {
     }
 }
 
-tasks.withType<JavaCompile>() {
+tasks.compileJava {
```
britter, line 23: This is not the same as before. Previously we were configuring all JavaCompile tasks. After this change we're just configuring the task with name `compileJava` but not `compileTestJava`. Whether or not this is what we want I can't say.

britter (commented):
Left a few improvement proposals. I jsut checked this from a stylistic point of view. I didn't check whether the Open API generator plugin configuration is correct. For that I would have to dig deep into this project.

components/capture-build-scan-url-maven-extension/build.gradle.kts
jthurne: > After this change we're just configuring the task with name compileJava but not compileTestJava

Thanks Benedikt! Do you know if `tasks.withType` is configuration-avoidance compatible? The docs about this particular method are somewhat ambiguous.

britter: I actually don't know. What I know is that `withType() {}` is not lazy, but `withType().configureEach {}` is. I think it's best to ask this question in #bt-questions.

sorin-florea (approved):

etiennestuder (approved):

PR #98, opened 2022-06-09
Use the Gradle Enterprise 2022.2.4 API to more accurately calculate executed cacheble task stats

jthurne:
Non-cacheable tasks that were non-cacheable because of overlapping outputs are
ripe for optimization (they can usually be fixed). Likewise, cacheable tasks
that were not cacheable in order to ensure correctness can also often be fixed
to be cacheable again. For these reasons, we want to treat such tasks as
cacheble tasks that were executed.

This change ensures the stats shown on the build scan page opened by the
_Executed cacheable tasks_ quick link match the stats shown in the experiment
summaries.

However, this change filters out such tasks from the stats for non-cachelable
tasks, which differs from build scan page that is opened via the _Executed
non-cacheable tasks_ quick link. The ideal solution would be to fix the quick
link to exlcude these tasks as well, but the current Gradle Enterprise UI
filtering options make this impossible.


jthurne:
This PR is a draft to prevent merging before PR #96 is merged.

etiennestuder (approved):

PR #105, opened 2022-06-18
Add support for configuring SSL when fetching build scan data

jthurne:
- Rename proxy.settings to network.settings
- Add support for allowing untrusted SSL certificates when fetching build scans
- Improve error handling for connection errors and workaround OpenAPI generator bug
- Only allow untrusted SSL certificates for the Gradle Enterprise host
- Respect the default trust store configured in the network.settings file


components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -157,6 +157,17 @@ public BuildValidationData fetchBuildValidationData(String buildScanId) {
                     throw new BuildScanNotFoundException(buildScanId, baseUrl, e.getCode(), e.getResponseBody(), e);
                 case StatusCodes.UNAUTHORIZED:
                     throw new AuthenticationFailedException(buildScanId, baseUrl, e.getCode(), e.getResponseBody(), e);
+                case 0:
+                    // This wasn't a bad HTTP response, instead it was a failure to connect.
+                    //
+                    // WORKAROUND: There is a bug in the OpenAPI generator that generates a broken ApiException.getMessage(),
```
marcphilipp, line 143: ❔ Is there an open issue for OpenAPI generator for this that we can link here?

components/scripts/maven/network.settings
```diff
@@ -1,16 +1,19 @@
...
 #https.proxyPassword=password
 
-# Uncomment to disable SSL certificate validation
+# Uncomment to disable SSL certificate validation for the Gradle Enterprise server
 #allowUntrustedServer=true
+
+# Uncomment and update to use a custom trust store
+#javax.net.ssl.trustStore=myTrustStore.jks
+#javax.net.ssl.trustStorePassword=changeit
+#javax.net.ssl.trustStoreType=jks
```
marcphilipp, line 19: ❓ Does OkHttp use these without any further configuration?

marcphilipp (approved):
Could we avoid duplication `network.settings` and copy it in the Gradle build instead? Should we call it `network.properties` for better editing support in IDEs?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
guylabs: It seems it's https://github.com/OpenAPITools/openapi-generator/issues/12631 and there should already be a fix for that on latest `master`.

components/scripts/maven/network.settings
jthurne: Yes. It took a bit of effort to test this, but I verified it works.

The `HandshakeCertificates.Builder.addPlatformTrustedCertificates` method fetches the JVM's default trust store and adds all of the certificates it finds in it to the TrustManager that it creates for use by OkHTTP.  The JVM's default trust store is influenced by these system properties.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
jthurne: Added a link to the issue to the comment.

jthurne:
> Could we avoid duplication network.settings and copy it in the Gradle build instead? 

✅ Done!

> Should we call it network.properties for better editing support in IDEs?

@etiennestuder what do you think?

If we do decide to rename it to `network.properties`, then we would probably want to also rename `mapping.example` to `mapping.properties.example` or something similar.

components/scripts/maven/mapping.example
```diff
@@ -1,9 +0,0 @@
-# This file serves as an example of how to define a custom mapping for the names of the custom values
-# that are extracted from the build scans that are fetched when running experiment 03 and 04.
-#
```
etiennestuder, line 3: By sharing the mapping file between Gradle and Maven, the text (which is currently build tool specific) has to become build tool agnostic.

```
...that are extracted from the build scans that are fetched when running the experiments that involve builds from CI.
```

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
```diff
@@ -45,8 +45,8 @@ public static class ExitCode {
     @Option(names = {"-m", "--mapping-file"}, description = "Specifies a mapping file that configures the keys used to fetch important custom values.")
```
etiennestuder, line 45: Do we need the `-m` option here, or is the `--mapping-file` option sufficient? I assume we always call this cmd line tool with a fixed arg name.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
```diff
@@ -45,8 +45,8 @@ public static class ExitCode {
...
     private Optional<Path> customValueMappingFile;
 
-    @Option(names = {"-p", "--proxy-settings-file"}, description = "Specifies a file that configures HTTP Proxy settings.")
-    private Optional<Path> proxySettingsFile;
+    @Option(names = {"--network-settings-file"}, description = "Specifies a file that configures HTTP Proxy and SSL settings.")
```
etiennestuder, line 48: `...that configures HTTP(S) Proxy and...`

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
```diff
@@ -59,7 +59,7 @@ public Integer call() throws Exception {
         // Use System.err for logging since we're going to write out the CSV to System.out
         logger = new ConsoleLogger(System.err, colorScheme, debug);
 
-        ProxyServerConfigurator.configureProxyServers(proxySettingsFile, logger);
+        NetworkingConfigurator.configureNetworking(networkSettingsFile, logger);
```
etiennestuder, line 62: In the code `networking` is used, for the file name `network` is used. Would it make sense to align on the naming? If you think so, too, then I would likely go with `network`. Curious what you think.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/ConsoleLogger.java
```diff
@@ -26,6 +26,10 @@ public void info(String message, Object... args) {
         out.printf(message, args);
     }
 
+    public void debug(String message, Object... args) {
```
etiennestuder, line 29: Line break missing after the `infoNoNewline` method.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/ConnectionFailedException.java
```diff
@@ -0,0 +1,15 @@
...
+import java.net.URL;
+
+public class ConnectionFailedException extends ApiClientException {
+
+    public ConnectionFailedException(String buildScanId, URL gradleEnterpriseServer, Throwable cause) {
+        super(String.format("Unable to connect to %s in order to fetch build scan %s: %s", gradleEnterpriseServer, buildScanId, cause.getMessage()), cause);
+    }
+
+    public ConnectionFailedException(String buildScanId, URL gradleEnterpriseServer, ApiException cause) {
+        super(String.format("Unable to connect to %s in order to fetch build scan %s.", gradleEnterpriseServer, buildScanId), cause);
```
etiennestuder, line 12: I would omit the dot at the end of the error message

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/ConnectionFailedException.java
```diff
@@ -0,0 +1,15 @@
+package com.gradle.enterprise.api.client;
+
+import java.net.URL;
+
+public class ConnectionFailedException extends ApiClientException {
+
+    public ConnectionFailedException(String buildScanId, URL gradleEnterpriseServer, Throwable cause) {
```
etiennestuder, line 7: For both constructors, I would swap the buildScanId and gradleEnterpriseServer parameters.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -37,33 +38,59 @@ public class GradleEnterpriseApiClient {
...
+        if (allowUntrustedServer()) {
+            trustedCertsBuilder.addInsecureHost(baseUrl.getHost());
+            httpClientBuilder.hostnameVerifier((hostname, session) -> baseUrl.getHost().equals(hostname));
+        }
+
+        HandshakeCertificates trustedCerts = trustedCertsBuilder.build();
+        httpClientBuilder.sslSocketFactory(trustedCerts.sslSocketFactory(), trustedCerts.trustManager());
+    }
+
+    private boolean allowUntrustedServer() {
```
etiennestuder, line 75: I would move this method below `configureProxyAuthentication`.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -37,33 +38,59 @@ public class GradleEnterpriseApiClient {
     private final ConsoleLogger logger;
 
     public GradleEnterpriseApiClient(URL baseUrl, CustomValueNames customValueNames, ConsoleLogger logger) {
+        this.logger = logger;
+        this.baseUrl = baseUrl;
         this.customValueNames = customValueNames;
+
```
etiennestuder, line 44: Assign them in the same order as passed in to the constructor.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -37,33 +38,59 @@ public class GradleEnterpriseApiClient {
...
+    private boolean allowUntrustedServer() {
+        return Boolean.parseBoolean(System.getProperty("allowUntrustedServer"));
+    }
+
+    private void configureProxyAuthentication(OkHttpClient.Builder httpClientBuilder) {
+        httpClientBuilder
+            .proxyAuthenticator((route, response) -> {
+                if (response.code() == 407) {
+                    String scheme = response.request().url().scheme().toLowerCase(Locale.ROOT);
```
etiennestuder, line 83: Is Local.ROOT env specific? If so, I would use a locale that is fixex and matches what a scheme can be encoded in. (ASCII?)

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -110,6 +137,18 @@ public BuildValidationData fetchBuildValidationData(String buildScanId) {
...
+                    //
+                    // WORKAROUND: There is a bug in the OpenAPI generator that generates an ApiException.getMessage() method
+                    // that throws a NullPointerException  when called on a non-HTTP error. See https://github.com/OpenAPITools/openapi-generator/issues/12631
+                    //
+                    // To work around the bug we use the underlying cause (if there is one). If there isn't a cause exception, then use
+                    // the ApiException but avoid calling its getMessage method.
+                    if (e.getCause() != null) {
+                        throw new ConnectionFailedException(buildScanId, baseUrl, e.getCause());
+                    }
+                    throw new ConnectionFailedException(buildScanId, baseUrl, e);
```
etiennestuder, line 151: I would use an `else` here to make it more explicit.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.Properties;
+
+public class NetworkingConfigurator {
+
+    public static void configureNetworking(Optional<Path> networkSettingsFile, ConsoleLogger logger) {
```
etiennestuder, line 14: I find an Optional as an arg not ideal. Instead, I would move that logic to the caller with something like networkSettingsFile.ifPresent(configure(it, logger)).

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+        networkSettingsFile.ifPresent(path -> {
+            try {
+                configureBasedOnProperties(path, logger);
+            } catch (IOException e) {
+                logger.debug("Unable to load settings from %s: %s", path, e.getMessage());
+                logger.debug(e);
+            }
```
etiennestuder, line 21: Should we throw here instead, given when this unlikely scenario happens, it is a serious issue that the user show know about?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+                .forEach(key -> System.setProperty(key, proxyProps.getProperty(key)));
+        }
+    }
+
+    private static boolean isNetworkingProperty(String key) {
+        return isSslProperty(key) || isProxyProperty(key);
+    }
+    private static boolean isSslProperty(String key) {
+        return key.startsWith("javax.net.ssl")
+            || key.equals("allowUntrustedServer");
```
etiennestuder, line 40: Should we change this property to `ssl.allowUntrustedServer`?

etiennestuder (changes requested):

components/scripts/maven/mapping.example
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
jthurne: Locale.ROOT is not env specific. It is the base Locale from which all other Locale's inherit from. Locale.ROOT is considered the best locale to use for locale insensitive strings (such as the URL scheme used here, which should always be lowercased the same way regardless of the user's locale).

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/ConnectionFailedException.java
jthurne: ✅ Done.  I did this for all of the other exceptions as well.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/ConnectionFailedException.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/ConsoleLogger.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
jthurne: We are currently using `-m`, but it's easy enough to change it, and I think that makes sense.

✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
jthurne: I renamed `NetworkingConfigurator` to `NetworkSettingsConfigurator`.  I went with "network settings" because I didn't want to imply that we're configuring the whole network for a host.  Although that still exists.

We could consider renaming everything to "HTTP settings" (`HttpConfigurator`, `http.settings`, etc), which might be slightly more accurate (all these settings only deal with configuring the HTTP).

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
jthurne: ✅ Done.

I went with `allowUntrustedServer` originally because that is what is used with the Gradle Enterprise plugin and Maven extension:

```
gradleEnterprise {
    server = "https://gradle-enterprise.mycompany.com"
    allowUntrustedServer = true
}
```

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
jthurne: Makes sense to me, ✅ Done.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/proxy/NetworkingConfigurator.java
jthurne: ✅ Done

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
etiennestuder: Interesting.

PR #120, opened 2022-07-22
Add --fail-if-not-fully-cacheable

jthurne:


components/scripts/lib/cli-parsers/common.m4
```diff
@@ -1,15 +1,15 @@
...
 # ARG_OPTIONAL_SINGLE([git-branch],[b],[])
 # ARG_OPTIONAL_SINGLE([git-commit-id],[c],[])
 # ARG_OPTIONAL_SINGLE([git-options],[o],[])
-# ARG_OPTIONAL_SINGLE([gradle-enterprise-server],[s],[],[])
-# ARG_OPTIONAL_SINGLE([git-repo],[r],[])
 # ARG_OPTIONAL_SINGLE([project-dir],[p],[],[])
+# ARG_OPTIONAL_SINGLE([args],[a],[])
+# ARG_OPTIONAL_BOOLEAN([enable-gradle-enterprise],[e],[])
 # ARG_OPTIONAL_BOOLEAN([interactive],[i],[],[off])
 # ARG_OPTIONAL_BOOLEAN([debug],[],[],[off])
```
etiennestuder, line 12: @jthurne why does it work that there are cmd line arguments defined in this common file that are not used by all scripts?

components/scripts/lib/cli-parsers/maven/03-cli-parser.m4
```diff
@@ -24,6 +24,7 @@ function print_help() {
   print_option_usage "-1, --first-build-ci" "Specifies the URL for the build scan of the first build run by a CI agent."
   print_option_usage "-2, --second-build-ci" "Specifies the URL for the build scan of the second build run by a CI agent."
   print_option_usage -m
+  print_option_usage -x
```
etiennestuder, line 27: It seems here we need to explicitly define `ARG_OPTIONAL_BOOLEAN([fail-if-not-optimized],[x],[])` since this script is not pulling in the _common_ script.

components/scripts/lib/cli-parsers/common.m4
```diff
@@ -1,15 +1,15 @@
...
-# ARG_OPTIONAL_SINGLE([args],[a],[])
 # ARGBASH_GO()
 # needed because of Argbash --> m4_ignore([
 ### START OF CODE GENERATED BY Argbash v2.10.0 one line above ###
```
etiennestuder, line 15: Could `ARG_VERSION` be moved to this script instead of being defined in each sub-script?

components/scripts/lib/cli-parsers/maven/03-cli-parser.m4
etiennestuder: Fixed.

components/scripts/lib/cli-parsers/common.m4
jthurne: I have a vague recollection of trying to put that here and it not working, but that was quite a long time ago. I think it's worth trying it again.

components/scripts/lib/cli-parsers/maven/03-cli-parser.m4
jthurne: Thanks! Good catch.

components/scripts/lib/cli-parsers/common.m4
jthurne: I just tried this, and unfortunately, moving `ARG_VERSION` to the common template breaks the version CLI argument. I also tried moving `ARG_HELP` up, but that broke the help CLI argument.  So unfortunately, these need to stay in the individual experiment CLI parser templates.

etiennestuder: Got it. Thanks for trying.

components/scripts/lib/help.sh
```diff
@@ -51,6 +51,9 @@ print_option_usage() {
     -v)
        _print_option_usage "-v, --version" "Prints version info."
        ;;
+    -x)
+       _print_option_usage "-x, --fail-if-not-optimized" "Causes the script to exit with exit code -2 if the script detects that the build is not optimized."
+       ;;
```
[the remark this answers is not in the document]
etiennestuder: I would use the following user-facing values/text:
- `-f`
- `--fail-if-not-fully-cacheable`
- "Terminates with exit code -2 if the build is not fully cacheable."
- FAILURE: Build is not fully cacheable for the given task graph. (with no line break after)

etiennestuder, line 56: I would use the following user-facing values/text:

-f
--fail-if-not-fully-cacheable
"Terminates with exit code -2 if the build is not fully cacheable."
FAILURE: Build is not fully cacheable for the given task graph. (with no line break after)

components/scripts/lib/info.sh
```diff
@@ -237,5 +237,16 @@ exit_with_return_code() {
...
   fi
 
+  # shellcheck disable=SC2034 # not all of the scripts have the fail if not optimized CLI argument
+  if [ -n "${_arg_fail_if_not_optimized+x}" ]; then
+    # If the script has the "fail if not optimized CLI argument", then see if the build is optimized
+    local executed_avoidable_tasks
+    executed_avoidable_tasks=$(( executed_cacheable_num_tasks[1] ))
+    if [[ "${_arg_fail_if_not_optimized}" == "on" ]] && (( executed_avoidable_tasks > 0 )); then
+      print_bl
+      die "ERROR: Build is not optimized" -2
```
etiennestuder, line 247: I'm not familiar on the convention for failure codes (negative, positive, etc.). We can discuss it.

components/scripts/lib/help.sh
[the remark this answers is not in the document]
jthurne: We return -1 if an unexpected error occurs while executing the scripts. Specifically, [the die function](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/6c370764031c8ace9292e909c3de6d085e48723b/components/scripts/lib/info.sh#L66-L72) will return -1 if a specific return code is not provided.

components/scripts/lib/info.sh
jthurne: I did some quick research on this, and it looks like it is more common to use positive exit codes instead of negative ones. Exit code 1 is usually reserved for "general failures" (like how we're currently using exit code -1). Otherwise, the exit codes are generally different across the board for different applications.

As such, I propose we use "1" for general failures, and "2" if the build is not fully cacheable.

components/scripts/lib/git.sh
```diff
@@ -29,8 +29,8 @@ git_clone_project() {
    debug git clone --depth=1 ${git_options} ${branch} "${git_repo}" "${clone_dir}"
 
    # shellcheck disable=SC2086  # we want $git_options and $branch to expand into multiple arguments
-   git clone --depth=1 ${git_options} ${branch} "${git_repo}" "${clone_dir}" || die "ERROR: Unable to clone from ${git_repo}." 2
-   cd "${clone_dir}" || die "Unable to access git repository directory ${clone_dir}." 2
+   git clone --depth=1 ${git_options} ${branch} "${git_repo}" "${clone_dir}" || die "ERROR: Unable to clone from ${git_repo}."
```
jthurne, line 32: @erichaagdev this change (and similar changes) will likely create a merge conflict between our two PRs, but it should be fairly easy to resolve.

jthurne, line 32: I decided it's better to not provide an explicit exit code here and instead rely on the default (which is `UNEXPECTED_ERROR`) because that pattern is easier to follow if we add additional `die` calls in the future.

components/scripts/lib/help.sh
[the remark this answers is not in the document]
jthurne: It turns out that we don't ever return -1. I was misinterpreting the code.  Anyway, it's all cleaned up now.

jthurne: ✅ Done

components/scripts/lib/libs.sh
```diff
@@ -1,37 +1,40 @@
 #!/usr/bin/env bash
 
 # shellcheck source=lib/color.sh
-source "${LIB_DIR}/color.sh" || { echo "Couldn't find '${LIB_DIR}/color.sh'"; exit 1; }
+source "${LIB_DIR}/color.sh" || { echo "Couldn't find '${LIB_DIR}/color.sh'"; exit 2; }
```
jthurne, line 4: Originally, I was thinking we can't use `die` here because it hasn't been loaded yet. And we can't load `exit-code.sh` first because it uses constants from `color.sh`.

Although, now that I've thought about it, I know how to work around this. `exit-code.sh` can load `color.sh` directly. It also depends on a function from `info.sh` (`print_ln`), but it could also load that directly. I'll give that a try tomorrow (and update this file to use `die`).

components/scripts/lib/exit-code.sh
```diff
@@ -0,0 +1,38 @@
...
+  fi
+  exit "${_ret}"
+}
+
+exit_with_return_code() {
+  if [[ " ${build_outcomes[*]} " =~ " FAILED " ]]; then
+    exit "${BUILD_FAILED}"
+  fi
+
+  # shellcheck disable=SC2034 # not all of the scripts have the fail if not optimized CLI argument
```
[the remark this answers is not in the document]
jthurne: ✅ Done in #168

components/scripts/lib/exit-code.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+SUCCESS=0
+INVALID_INPUT=1
+UNEXPECTED_ERROR=2
+BUILD_FAILED=3
+BUILD_IS_NOT_FULLY_CACHEABLE=4
```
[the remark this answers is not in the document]
jthurne: ✅ Done in #168

etiennestuder:
Very thorough and reproducible testing. Thanks, @erichaagdev.

PR #133, opened 2022-08-19
Replace reference to local build cache with incremental build reference

### LINKED ISSUE #132: Gradle 01-validate-incremental-building.sh should reference "up-to-date checks" instead of "local build cache"
Text in `01-validate-incremental-building.sh` currently reads:

> ...how well your build leverages Gradle’s local build cache for the invoked set of Gradle tasks

"local build cache" should say "up-to-date checks" or "incremental build" instead because that's the topic of the experiment.

jthurne:
Fixes #132

### DIFF
--- a/components/scripts/gradle/01-validate-incremental-building.sh
+++ b/components/scripts/gradle/01-validate-incremental-building.sh
@@ -219,8 +219,8 @@ $(print_separator)
 ${HEADER_COLOR}Measure build results${RESTORE}
 
 Now that the second build has finished successfully, you are ready to measure in
-Gradle Enterprise how well your build leverages Gradle’s local build cache for
-the invoked set of Gradle tasks.
+Gradle Enterprise how well your build leverages Gradle’s incremental building
+functionality for the invoked set of Gradle tasks.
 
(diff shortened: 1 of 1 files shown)


components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -219,8 +219,8 @@ $(print_separator)
 ${HEADER_COLOR}Measure build results${RESTORE}
 
 Now that the second build has finished successfully, you are ready to measure in
-Gradle Enterprise how well your build leverages Gradle’s local build cache for
-the invoked set of Gradle tasks.
+Gradle Enterprise how well your build leverages Gradle’s incremental building
```
runningcode, line 222:
```suggestion
Gradle Enterprise how well your build leverages Gradle’s incremental build
```

This is referred to simply as `incremental build` in the docs. I would drop the word `functionality`.

IMO, to be more clear, it would be best to use `up-to-date` instead of `incremental` since this would make it more clear that this differs from incremental task execution.

etiennestuder:
Nice catch. Thanks.

Issue #148, opened 2022-09-27
Experiment 5 provides incorrect summary for avoided / executed cacheable tasks


erichaagdev:
When running experiment 5 (Gradle), I'm seeing an incorrect report for "Avoided cacheable tasks" and "Executed cacheable tasks".

![image](https://user-images.githubusercontent.com/5797900/192579953-2981577e-8130-439e-b111-38c44240c74b.png)

This is different than what is shown in the actual Build Scan results.

![image](https://user-images.githubusercontent.com/5797900/192580054-3b9c5efb-b0fd-48e8-bd9f-3a6c308b835e.png)

As you can see, the validation script summary indicates cacheable tasks were executed, while the Build Scan results indicate the opposite. 

Additional information:
- Using the latest 2.0 release of the validation scripts

etiennestuder:
Thanks for the fix @jthurne , and thanks for raising this, @erichaagdev.

PR #149, opened 2022-09-27
Fix build caching stats for Gradle exp 05/Maven exp 04

jthurne:
Previously, the arrays containing the build scan data were built using append operations. Since Gradle Exp 05 and Maven Exp 04 fetch build scans twice (once to get the build parameters for the 2nd build), this means the data for the first build was appended to the array twice. When it came time to display the statistics, the scripts always show the data at array index 1 (zero-based indexing), so the statistics for the first build would be displayed.

See #148

### DIFF
--- a/components/scripts/lib/build_scan.sh
+++ b/components/scripts/lib/build_scan.sh
@@ -188,6 +188,7 @@ fetch_and_read_build_scan_data() {
 
   header_row_read=false
   idx=0
+  
   # shellcheck disable=SC2034 # not all scripts use all of the fetched data
   while IFS=, read -r field_1 field_2 field_3 field_4 field_5 field_6 field_7 field_8 field_9 field_10 field_11 field_12 field_13 field_14 field_15 field_16 field_17 field_18 field_19; do
      if [[ "$header_row_read" == "false" ]]; then
--- a/components/scripts/lib/build_scan.sh
+++ b/components/scripts/lib/build_scan.sh
@@ -154,7 +154,7 @@ fi
 fetch_and_read_build_scan_data() {
   # This isn't the most robust way to read a CSV,
   # but we control the CSV so we don't have to worry about various CSV edge cases
-  local args build_cache_metrics_only fetched_data header_row_read
+  local args build_cache_metrics_only fetched_data header_row_read idx
   args=()
 
   if [[ "$_arg_debug" == "on" ]]; then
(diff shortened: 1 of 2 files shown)


erichaagdev:
I re-ran experiment 5 with these changes and I can confirm this resolves the issue.

image

Build Scan: https://ge.solutions-team.gradle.com/s/ut4ojmqgjck2s/performance/execution

Nice work, @jthurne!!

erichaagdev (approved):

components/scripts/lib/build_scan.sh
```diff
@@ -187,6 +187,8 @@ fetch_and_read_build_scan_data() {
   debug ""
 
   header_row_read=false
+  idx=0
```
etiennestuder, line 190: Will the value of idx ever be something other than 0?

components/scripts/lib/build_scan.sh
```diff
@@ -154,7 +154,7 @@ fi
 fetch_and_read_build_scan_data() {
   # This isn't the most robust way to read a CSV,
   # but we control the CSV so we don't have to worry about various CSV edge cases
-  local args build_cache_metrics_only fetched_data header_row_read
+  local args build_cache_metrics_only fetched_data header_row_read idx
```
etiennestuder, line 157: What does it mean to pass `idx` here?

components/scripts/lib/build_scan.sh
jthurne: This just initializes the variable to 0. It's incremented later on in the loop.  If we don't initialize the variable to zero, then I'm not sure what the behavior will be, as it will have an undefined value when we increment it.

components/scripts/lib/build_scan.sh
jthurne: This declares that `idx` is a local variable to this function. So after this function is done executing, it won't be accessible anymore.

Unlike other languages, which assume variables are local unless otherwise specified, bash assumes all variables are global (unless otherwise specified).

components/scripts/lib/build_scan.sh
etiennestuder: I had overlooked the incrementing in the loop. All clear now. Thanks, @jthurne.

components/scripts/lib/build_scan.sh
etiennestuder: Got it. Thanks for explaining, @jthurne.

PR #150, opened 2022-09-27
Upgrade Java toolchain version from 8 to 11

erichaagdev:
On an M1 MacBook Pro, a build will fail if a local installation of JDK 8 does not exist. This is due to the fact that Adoptium [does not provide a compatible version of JDK 8](https://adoptium.net/temurin/releases?version=8) for `aarch64`. 

The purpose of this PR is to workaround this by temporarily upgrading the Java toolchain version to 11 while we wait for [a Gradle Build Tool fix](https://github.com/gradle/gradle/issues/19140), at which point, this change can be reverted.

See also: gradle/gradle-build-scan-quickstart#32

### DIFF
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -13,10 +13,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK 8
         uses: actions/setup-java@v2
         with:
-          java-version: '11'
+          java-version: '8'
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -13,10 +13,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@v2
         with:
-          java-version: '8'
+          java-version: '11'
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
--- a/.github/workflows/development-release.yml
+++ b/.github/workflows/development-release.yml
@@ -17,10 +17,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK 8
         uses: actions/setup-java@v2
         with:
-          java-version: '11'
+          java-version: '8'
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
--- a/.github/workflows/development-release.yml
+++ b/.github/workflows/development-release.yml
@@ -17,10 +17,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Set up JDK 8
+      - name: Set up JDK 11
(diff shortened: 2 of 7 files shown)


etiennestuder:
I don't understand why the yaml files need to be changed, too. The idea of the toolchain is that one can decouple the Java version used to run the Gradle build from the Java version used to compile, test, etc.

etiennestuder:
Also, if we use Java 11 in the toolchain, it means the Java classes we create will require to also use Java 11 to run the scripts (which invoke the java cmd line utilities).

etiennestuder:
> On an M1 MacBook Pro, a build will fail if a local installation of JDK 8 does not exist. This is due to the fact that Adoptium [does not provide a compatible version of JDK 8](https://adoptium.net/temurin/releases?version=8) for `aarch64`.

I don't quite follow. The link you provide points to Temurin. And there, there is a JDK8 version for aarch64:
https://adoptium.net/de/temurin/releases/?version=8

image

erichaagdev:
> I don't quite follow. The link you provide points to Temurin. And there, there is a JDK8 version for aarch64:
https://adoptium.net/de/temurin/releases/?version=8

That JDK is for Linux, not macOS. If you switch the [JDK version to 11](https://adoptium.net/de/temurin/releases/?version=11), you can see there is an explicit entry for macOS.

jthurne:
Could we make this change conditional?  Use the Java 8 toolchain, unless the build is running on an Apple Silicone Macbook, in which case use the Java 11 toolchain?

Alternatively, if we use the Java 11 toolchain, can we still configure the build to use Java 8 source level, and produce Java 8 bytecode?

erichaagdev:
> Alternatively, if we use the Java 11 toolchain, can we still configure the build to use Java 8 source level, and produce Java 8 bytecode?

I've pushed changes that I believe does exactly this. When running locally, a compatible version of JDK 11 is auto provisioned and I can run `fetch-build-scan-data-cmdline-tool` using an already existing installation of Java 8.

image

@etiennestuder, please let me know if you feel this is an acceptable solution. With this approach I think we should configure the job to use Java 11 to avoid the need to auto provision it. 

Note: I think CI may be failing due to a bad build cache key. I ran a build with no changes aside from [adding a build cache shard](https://ge.solutions-team.gradle.com/s/7ne2zemf5jcmk/performance/build-cache#remote-cache-url) and achieved a successful build. [Without the shard](https://ge.solutions-team.gradle.com/s/4kcswkipdygjk/performance/build-cache#remote-cache-url), the build fails.

runningcode:
I think there are several confusing factors here but hopefully this should clear it up.

From a high level, we want these scripts to always be compatible with JDK8. This is why the toolchain declaration exists. We shouldn't change this. It is not clear to me that your change Eric does this. It isn't clear to me what the behavior on the `JavaCompile` task would be. Would the java release be overridden by the toolchain? The tests will now be run using java 11 which is not what we want.

The fact that the Gradle build tool does not automatically download a compatible JDK in the java 8 mac arm64 case is a separate issue. To resolve that issue, you can download a compatible JDK for your machine. I recommend [Zulu](https://www.azul.com/downloads/?version=java-8-lts&os=macos&architecture=arm-64-bit&package=jdk).
You can also configure the toolchain location for your machine [like so](https://docs.gradle.org/current/userguide/toolchains.html#sec:custom_loc).

runningcode (changes requested):
I propose closing this PR.

erichaagdev:
> It is not clear to me that your change Eric does this. It isn't clear to me what the behavior on the JavaCompile task would be.

I agree with this. Closing this PR.

PR #156, opened 2022-10-12
Append a trailing slash to the build cache URL

### LINKED ISSUE #155: Experiment 5 fails when using GE Build Cache connector due to missing trailing slash
When running experiment 5 using Gradle Enterprise 2022.3.5, experiment 5 fails due to the build cache URL returned from the Gradle Enterprise API not having a trailing `/`.

image

erichaagdev:
This change appends a `/` to the remote build cache URL path if one does not exist in the `configure-remote-build-caching` init script.

### Before

image

### After

image

The remote build cache URL supplied to the build is unchanged from what is returned in the API (note the **_missing_** `/`).

image

Fixes #155

### DIFF
--- a/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
@@ -24,7 +24,11 @@ if (isTopLevelBuild) {
 
 URI getRemoteBuildCacheUrl() {
     def projectProperties = gradle.startParameter.projectProperties
-    new URI(projectProperties.get("com.gradle.enterprise.build_validation.remoteBuildCacheUrl"))
+    if (projectProperties.containsKey("com.gradle.enterprise.build_validation.remoteBuildCacheUrl")) {
+        return new URI(projectProperties.get("com.gradle.enterprise.build_validation.remoteBuildCacheUrl"))
+    } else {
+        return null
+    }
 }
 
 static URI withPathTrailingSlash(URI uri) {
--- a/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
@@ -9,13 +9,9 @@ if (isTopLevelBuild) {
             remote(HttpBuildCache) {
                 enabled = true
                 push = false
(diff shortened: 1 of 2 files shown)


runningcode:
Just for my understanding if the user has their build configured incorrectly, will this append the `/` to their build? Or is this only for build cache URL's passed in via arguments to the scripts?

erichaagdev:
> Or is this only for build cache URL's passed in via arguments to the scripts?

Correct. Take a look at #155 if you haven't already.

jthurne (approved):

etiennestuder:
IMO, this change is OK. What is not great is that if someone is using the new connector, we will also _display_ the cache URL with a trailing slash in the summary. Wouldn't it be better to append the slash at the time where we _apply_ instead of at the time we _read_ this value?

jthurne:
> IMO, this change is OK. What is not great is that if someone is using the new connector, we will also _display_ the cache URL with a trailing slash in the summary. Wouldn't it be better to append the slash at the time where we _apply_ instead of at the time we _read_ this value?

Taking this approach instead sounds fine to me. 👍

erichaagdev:
@etiennestuder and @jthurne I have updated the PR and description based on these suggestions.

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
```diff
@@ -9,11 +9,24 @@ if (isTopLevelBuild) {
...
+                    }
                 }
             }
         }
     }
 }
+
+URI getRemoteBuildCacheUrl() {
+    def projectProperties = gradle.startParameter.projectProperties
+    new URI(projectProperties.get("com.gradle.enterprise.build_validation.remoteBuildCacheUrl"))
```
jthurne, line 27: I believe this will generate an NPE if the project property is missing.  And it looks like we checked to see if the project property existed before reconfiguring build caching.

So we may want to update this to check to see if the project property exists, and if not, return a placeholder (maybe null), so that the if statement on line 13 above will do what is intended.

erichaagdev: Nice catch. Addressed by 8d5237eb3a7924c296757edf233ed0d46b013e77

jthurne (approved):

Issue #138, opened 2022-09-08
Scripts can fail on projects that require git history


clayburn:
We've seen some cases where git logic in tasks or build scripts can cause the scripts to fail due to the shallow cloning of the git repository. For example, a git task that searches for the most recent tag on a given branch can fail when the repo is cloned with a depth of 1.

In most cases, the shallow clone works great, so I believe it should remain the default. However, it may be nice to have a flag to grab a full clone in cases where this is an issue.

etiennestuder:
The coming version of the build validation scripts will allow to configure Git options used when cloning the repo. See [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/166).

//cc @clayburn

PR #164, opened 2022-10-25
Add `-o`/`--git-options`

erichaagdev:
This PR adds the ability to supply additional options the `git clone` by use  of the `--git-options` parameter, or `-o` for short.

Example:

```shell
./01-validate-incremental-building.sh  --git-options '--depth=9999' --tasks build -r 'git@github.com:gradle/gradle-enterprise-build-validation-scripts.git'
```

Experiments affected:
- Gradle: 1, 2, 3, & 5
- Maven: 1, 2, & 4

Notes/questions:
- When checking out a specific commit, which occurs when `--git-commit` is used or always in gradle-exp5 and maven-exp4(?), we `fetch` instead of a `clone`. Should we also supply the value of `--git-options` to `fetch`?
- The common case of wanting to perform a full clone is not directly supported by this change. We are already using `--depth=1` when we `clone` with no way to remove it. A workaround for this is to call the script with `--git-options '--depth=99999'`.

### DIFF
--- a/components/scripts/lib/cli-parsers/common.m4
+++ b/components/scripts/lib/cli-parsers/common.m4
@@ -3,6 +3,7 @@
 # ARG_OPTIONAL_BOOLEAN([enable-gradle-enterprise],[e],[])
 # ARG_OPTIONAL_SINGLE([git-branch],[b],[])
 # ARG_OPTIONAL_SINGLE([git-commit-id],[c],[])
+# ARG_OPTIONAL_SINGLE([git-options],[o],[])
 # ARG_OPTIONAL_SINGLE([gradle-enterprise-server],[s],[],[])
 # ARG_OPTIONAL_SINGLE([git-repo],[r],[])
 # ARG_OPTIONAL_SINGLE([project-dir],[p],[],[])
--- a/components/scripts/lib/config.sh
+++ b/components/scripts/lib/config.sh
@@ -149,6 +149,7 @@ collect_git_commit_id() {
 collect_git_options() {
    local default_git_options="<none>"
    prompt_for_setting "What are additional options to use when cloning the Git repository?" "${git_options}" "${default_git_options}" git_options
+   
    if [[ "${git_options}" == "${default_git_options}" ]]; then
      git_options=''
    fi
--- a/components/scripts/lib/config.sh
+++ b/components/scripts/lib/config.sh
@@ -17,6 +17,10 @@ process_arguments() {
     git_commit_id="${_arg_git_commit_id}"
   fi
 
+  if [ -n "${_arg_git_options+x}" ]; then
+    git_options="${_arg_git_options}"
+  fi
+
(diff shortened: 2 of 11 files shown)


components/scripts/lib/config.sh
```diff
@@ -116,6 +120,7 @@ collect_git_details() {
   collect_git_repo
   collect_git_branch
   collect_git_commit_id
+  collect_git_options
```
jthurne, line 123: I had not thought about asking for this in interactive mode. 🤔 . I could go either way with this here. On the one hand, we ask for additional arguments to pass to Gradle/Maven, and this is similar. On the other hand, it should be uncommon for a user to set `--git-options`, so I don't know how important it is to ask for it in interactive mode.

I'll be interested in what Etienne thinks.

jthurne: Nevermind, I see that Etienne asked for this already. 👍

erichaagdev:
Should this close #138?

components/scripts/lib/cli-parsers/common.m4
```diff
@@ -3,6 +3,7 @@
 # ARG_OPTIONAL_BOOLEAN([enable-gradle-enterprise],[e],[])
 # ARG_OPTIONAL_SINGLE([git-branch],[b],[])
 # ARG_OPTIONAL_SINGLE([git-commit-id],[c],[])
+# ARG_OPTIONAL_SINGLE([git-options],[o],[])
```
etiennestuder, line 6: In interactive mode, we need to print the `-o` value in the end-of-experiment summary if it differs from the default. Currently, we don't.

image

etiennestuder (changes requested):
@erichaagdev Thanks for the PR. I've left two comments to be addressed.

components/scripts/lib/cli-parsers/common.m4
etiennestuder, line 6: We also need to print the options in the summary section.

image

components/scripts/lib/git.sh
```diff
@@ -24,8 +24,12 @@ git_clone_project() {
    fi
 
    rm -rf "${clone_dir:?}"
-   # shellcheck disable=SC2086  # we want $branch to expand into multiple arguments
-   git clone --depth=1 ${branch} "${git_repo}" "${clone_dir}" || die "ERROR: Unable to clone from ${git_repo}." 2
+
+   # shellcheck disable=SC2086  # we want $git_options and $branch to expand into multiple arguments
+   debug git clone --depth=1 ${git_options} ${branch} "${git_repo}" "${clone_dir}"
```
etiennestuder, line 29: I suggest we also add `--shallow-submodules` by default.

jthurne: +1. I was thinking this yesterday as well after I reviewed the `git clone --help` content.

PR #168, opened 2022-10-27
Use sane exit codes and use them consistently

jthurne:
Previously, the exit codes we returned were all over the place. THis commit cleans them up by:

  * Introducing a set of constants for the conditions we want to return a different exit code.
  * Only use positive exit codes

See the new constants in exit_code.sh for what the new exit codes are.

Also ensures a newline is only printed after an error message if the CLI help is displayed.


components/scripts/lib/exit-code.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+SUCCESS=0
+INVALID_INPUT=1
+UNEXPECTED_ERROR=2
```
jthurne, line 5: I tried swapping 1 and 2, but unfortunately, the code argbash generates always uses exit code 1 for unknown arguments (or for any argument validation).  I'll keep digging a bit to see if that can be influenced or changed.

etiennestuder: Got it. I would not break a leg over this.

components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
```diff
@@ -137,7 +137,7 @@ parse_build_scan_urls() {
       base_urls+=("${protocol}://${ge_host}${port}")
       build_scan_ids+=("$build_scan_id")
     else
-      die "${url} is not a parsable URL." 4
+      die "${url} is not a parsable URL."
```
etiennestuder, line 140: @jthurne should we return "${INVALID_INPUT}" too here?

components/scripts/lib/build_scan.sh
```diff
@@ -72,14 +72,14 @@ fetch_build_scan_data() {
           die "ERROR: JAVA_HOME is set to an invalid directory: $JAVA_HOME
 
   Please set the JAVA_HOME variable in your environment to match the
-  location of your Java installation."
+  location of your Java installation." "${INVALID_INPUT}"
```
etiennestuder, line 75: Not sure `INVALID_INPUT` is right here.

components/scripts/lib/build_scan.sh
```diff
@@ -72,14 +72,14 @@ fetch_build_scan_data() {
...
       fi
   else
       JAVACMD="java"
       which java >/dev/null 2>&1 || die "ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH.
 
   Please set the JAVA_HOME variable in your environment to match the
-  location of your Java installation."
+  location of your Java installation." "${INVALID_INPUT}"
```
etiennestuder, line 82: Not sure `INVALID_INPUT` is right here.

components/scripts/lib/libs.sh
```diff
@@ -1,37 +1,40 @@
...
-source "${LIB_DIR}/help.sh" || { echo "Couldn't find '${LIB_DIR}/help.sh'"; exit 1; }
+source "${LIB_DIR}/help.sh" || { echo "Couldn't find '${LIB_DIR}/help.sh'"; exit 2; }
 
 # shellcheck source=lib/paths.sh
-source "${LIB_DIR}/paths.sh" || { echo "Couldn't find '${LIB_DIR}/paths.sh'"; exit 1; }
+source "${LIB_DIR}/paths.sh" || { echo "Couldn't find '${LIB_DIR}/paths.sh'"; exit 2; }
 
 # shellcheck source=lib/project.sh
-source "${LIB_DIR}/project.sh" || { echo "Couldn't find '${LIB_DIR}/project.sh'"; exit 1; }
+source "${LIB_DIR}/project.sh" || { echo "Couldn't find '${LIB_DIR}/project.sh'"; exit 2; }
```
etiennestuder, line 40: Shouldn't we return a constant instead of 2 in all these cases here and above?

components/scripts/lib/maven.sh
```diff
@@ -14,11 +14,11 @@ invoke_maven() {
   local args mvn
   args=()
 
-  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." 3
+  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." "${INVALID_INPUT}"
 
   mvn=$(find_maven_executable)
   if [ -z "$mvn" ]; then
-    die "Unable to find the Maven executable. Add MAVEN_INSTALL_DIR/bin to your PATH environment variable, or install the Maven Wrapper." 2
+    die "Unable to find the Maven executable. Add MAVEN_INSTALL_DIR/bin to your PATH environment variable, or install the Maven Wrapper." "${INVALID_INPUT}"
```
etiennestuder, line 21: Not sure `INVALID_INPUT` is right here.

components/scripts/lib/git.sh
```diff
@@ -56,9 +56,9 @@ git_checkout_commit() {
 
   rm -rf "${clone_dir:?}"
   mkdir -p "${clone_dir}"
-  cd "${clone_dir}" || die "ERROR: Unable access git repository directory ${clone_dir}" 2
-  git init > /dev/null || die "ERROR: Unable initialize git" 2
-  git remote add origin "${git_repo}" || die "ERROR: Unable to fetch from ${git_repo}" 2
+  cd "${clone_dir}" || die "ERROR: Unable access git repository directory ${clone_dir}"
```
etiennestuder, line 59: Unable `to` access...

components/scripts/maven/03-validate-remote-build-caching-ci-ci.sh
```diff
@@ -135,7 +135,7 @@ parse_build_scan_urls() {
       base_urls+=("${protocol}://${ge_host}${port}")
       build_scan_ids+=("$build_scan_id")
     else
-      die "${url} is not a parsable URL." 4
+      die "${url} is not a parsable URL."
```
etiennestuder, line 138: Should we use `INVALID_INPUT` here.

etiennestuder (changes requested):

components/scripts/lib/libs.sh
jthurne: Yes, however, the constants aren't loaded until the `exit-code.sh` script is loaded.  That said, we could define a constant in this file, for this file.  That would still be more readable. I'll make that change.

jthurne: I pushed up a commit that cleans this up significantly, and also ensures that the error printed if failing to load a library is consistently formatted.

components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -22,9 +22,9 @@ readonly LIB_DIR="${SCRIPT_DIR}/lib"
 
 # Include and parse the command line arguments
 # shellcheck source=lib/01-cli-parser.sh
-source "${LIB_DIR}/${EXP_NO}-cli-parser.sh" || { echo "Couldn't find '${LIB_DIR}/${EXP_NO}-cli-parser.sh' library."; exit 1; }
+source "${LIB_DIR}/${EXP_NO}-cli-parser.sh" || { echo -e "\033[00;31m\033[1mERROR: Couldn't find '${LIB_DIR}/${EXP_NO}-cli-parser.sh' parsing library.\033[0m"; exit 2; }
```
jthurne, line 25: The funny looking `\033[00;` bits are to change the text color to red, bold the text, and then reset the formatting to the terminal's default after the error message has been printed.

Elsewhere we hide this with functions and constants, but since we haven't loaded those functions and constants yet, we can't rely on them here. Same reason why the exit code is hardcoded rather than using one of our constants.

components/scripts/lib/libs.sh
```diff
@@ -1,37 +1,52 @@
 #!/usr/bin/env bash
 
+failed_to_load_lib() {
+  local lib_name _RED _BOLD _RESTORE _UNEXPECTED_ERROR
```
jthurne, line 4: Have to prefix the local constants here with `_` so that they don't conflict with the other constants that are loaded by the color.sh and exit-code.sh library scripts.

components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
jthurne: Yes, you're right. Will fix.

components/scripts/maven/03-validate-remote-build-caching-ci-ci.sh
jthurne: ✅ fixed

components/scripts/lib/git.sh
jthurne: ✅ fixed

components/scripts/lib/maven.sh
jthurne: ✅ Removed

components/scripts/lib/build_scan.sh
jthurne: ✅ Removed

components/scripts/lib/build_scan.sh
jthurne: ✅ Removed

components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
jthurne: ✅ fixed

components/scripts/lib/gradle.sh
```diff
@@ -13,7 +13,7 @@ invoke_gradle() {
   local init_scripts_dir
   init_scripts_dir="$(init_scripts_path)"
 
-  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." 3
+  pushd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}."
```
etiennestuder, line 16: For maven, this is defined as a INVALID_INPUT

etiennestuder (changes requested):

components/scripts/lib/exit-code.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+SUCCESS=0
+INVALID_INPUT=1
+BUILD_FAILED=3
+BUILD_NOT_FULLY_CACHEABLE=4
+UNEXPECTED_ERROR=100
+
```
etiennestuder, line 8: SHould be 0, 1, 2, 3,...

components/scripts/lib/exit-code.sh
```diff
@@ -0,0 +1,37 @@
...
+exit_with_return_code() {
+  if [[ " ${build_outcomes[*]} " =~ " FAILED " ]]; then
+    exit "${BUILD_FAILED}"
+  fi
+
+  # shellcheck disable=SC2034 # not all of the scripts have the --fail-if-not-fully-cacheable CLI argument
+  if [ -n "${fail_if_not_fully_cacheable+x}" ]; then
+    local executed_avoidable_tasks
+    executed_avoidable_tasks=$(( executed_cacheable_num_tasks[1] ))
+    if [[ "${fail_if_not_fully_cacheable}" == "on" ]] && (( executed_avoidable_tasks > 0 )); then
```
etiennestuder, line 31: "move up two lines".

etiennestuder:
@jthurne Please merge to master once remaining comments have been addressed.

components/scripts/lib/gradle.sh
jthurne: ✅ Done

components/scripts/lib/exit-code.sh
jthurne: ✅ Done

components/scripts/lib/exit-code.sh
jthurne: ✅ Done

PR #174, opened 2022-10-28
Add offline flag to Gradle experiment 3

erichaagdev:


components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -38,6 +38,10 @@ ge_server=''
 interactive_mode=''
 
 main() {
+  if [[ "$_arg_debug" == "on" && "$offline_mode" == "on" ]]; then
+    debug "Running experiment in offline mode."
+  fi
+
```
etiennestuder, line 44: I'm not sure about the relationship established here between debug and offline.

components/scripts/lib/config.sh
```diff
@@ -62,6 +62,10 @@ process_arguments() {
   if [ -n "${_arg_interactive+x}" ]; then
     interactive_mode="${_arg_interactive}"
   fi
+
+  if [ -n "${_arg_offline+x}" ]; then
+    offline_mode="${_arg_offline}"
+  fi
```
etiennestuder, line 68: do you know why `debug` is not processed in this method?

components/scripts/lib/config.sh
```diff
@@ -62,6 +62,10 @@ process_arguments() {
   if [ -n "${_arg_interactive+x}" ]; then
     interactive_mode="${_arg_interactive}"
   fi
+
+  if [ -n "${_arg_offline+x}" ]; then
```
etiennestuder, line 66: I would move this if-block above the interactive if-block.

etiennestuder (approved):

components/scripts/lib/config.sh
jthurne: We can't use a variable named `debug` because we also have a function called `debug`, so I was just using the `_arg_debug` variable directly.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
jthurne: This could be simplified by removing the debug check. The debug function does that check already.

components/scripts/lib/config.sh
erichaagdev: What about `debug_mode`? We use the `x_mode` convention in a few places now.

jthurne: Yeah, `debug_mode` would work.

jthurne: We probably want to change that in a different PR though.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Removed the redundant debug check in 77a18f5527cc5fba4c16e59daa112563a5e74bd8.

components/scripts/lib/config.sh
erichaagdev: Addressed in c9110cc807f3057942e957a0cd8f3a9066006701.

jthurne (approved):

components/scripts/lib/config.sh
etiennestuder: I've created a separate [issue](https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/180).

components/scripts/lib/config.sh
```diff
@@ -268,6 +272,10 @@ print_command_to_repeat_experiment() {
     cmd+=("-f")
   fi
 
+  if [[ "${offline_mode}" == "on" ]]; then
+    cmd+=("--offline")
+  fi
+
```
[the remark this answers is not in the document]
etiennestuder: I've created a separate [issue](https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/181).

PR #177, opened 2022-10-28
Fix --no-build-scan-publishing

jthurne:
The `--no` prefix was confusing Argbash, which supports turning boolean flags
off with `--no-` prefixes.  The solution is to define a `build-scan-publishing`
flag instead, but default it to 'on'.

### DIFF
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -38,7 +38,7 @@ ge_server=''
 interactive_mode=''
 
 main() {
-  if [[ "$no_build_scan_publishing_mode" == "on" ]]; then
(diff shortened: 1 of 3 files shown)


components/scripts/lib/info.sh
```diff
@@ -101,8 +101,11 @@ print_summary() {
   print_build_scans
   print_warnings
   print_performance_metrics
-  print_bl
-  print_quick_links
+
+  if [[ "${build_scan_publishing_mode}" == "on" ]]; then
+    print_bl
+    print_quick_links
+  fi
```
jthurne, line 108: I didn't mean to push this up yet. Removing.

jthurne: All better.

erichaagdev (approved):

PR #179, opened 2022-10-29
Use Build Scan dumps when publishing is disabled for Gradle experiment 3

### LINKED ISSUE #170: Gradle experiment 3 can be run without Build Scan publishing
We should provide a mechanism to run experiments without Build Scan publishing for cases where the Gradle Enterprise server may not yet be available.

### LINKED ISSUE #184: Build does not try to publish build scan when publishing has been disabled on the experiment
- we still need to disable build scan publishing when the cmd line flag is set accordingly. This should be quite forward since we already have a place where the script currently tells the build that it should always publish a build scan

image

- we also need to adjust the cmd line output shown to the user (should have no --scans, but the system property instead). this might also affect the debug info shown.

image

(see original [comment](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/178#issuecomment-1296256047))

erichaagdev:
### How to test this

1. Checkout this branch `erichaagdev/parse-scan-dump`
2. Run `./gradlew clean build`
3. `cd build/distributions`
4. `unzip gradle-enterprise-gradle-build-validation-dev.zip`
5. `cd gradle-enterprise-gradle-build-validation`
6. Download a copy of the `build-scan-support-tool.jar` and place it in the current directory. It must be named exactly `build-scan-support-tool.jar`.
7. Run! 

```shell
./03-validate-local-build-caching-different-locations.sh \
  --git-repo 'https://github.com/etiennestuder/java-ordered-properties' \
  --git-branch ge \
  --tasks build \
  --disable-build-scan-publishing
```

Here is an example of what it looks like when the `build-scan-support-tool.jar` is not found in the same directory as the validation scripts.

image

Additionally, the message "Extracting build scan data, done." has been added to display the status of the Build Scan extraction.

image

Finally, the Gradle command displayed to the user changes conditionally based on whether `--disable-build-scan-publishing` is used.

**With `--disable-build-scan-publishing`**

image

**Without `--disable-build-scan-publishing`**

image

Closes #170
Closes #184

### DIFF
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -19,6 +19,7 @@ readonly SCRIPT_NAME
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo .)")"; pwd)"
 readonly SCRIPT_DIR
 readonly LIB_DIR="${SCRIPT_DIR}/lib"
+readonly SCANS_SUPPORT_TOOLS_JAR="${SCRIPT_DIR}/scans-support-tools.jar"
 
 # Include and parse the command line arguments
 # shellcheck source=lib/03-cli-parser.sh
@@ -38,6 +39,10 @@ ge_server=''
 interactive_mode=''
 
 main() {
+  if [[ "$build_scan_publishing_mode" == "off" ]]; then
+    verify_scans_support_tools
+  fi
+
   if [ "${interactive_mode}" == "on" ]; then
     wizard_execute
   else
@@ -45,6 +50,12 @@ main() {
   fi
   create_receipt_file
   exit_with_return_code
+}
+
+verify_scans_support_tools() {
+  if [ ! -f "$SCANS_SUPPORT_TOOLS_JAR" ]; then
+    die "ERROR: scans-support-tools.jar is required when running with --no-build-scan-publishing."
+  fi
 }
 
 execute() {
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -38,6 +38,10 @@ enable_ge=''
 ge_server=''
 interactive_mode=''
 
+# Used with Build Scan publishing disabled
+first_build_scan_dump=''
+second_build_scan_dump=''
+
 main() {
   if [[ "$build_scan_publishing_mode" == "off" ]]; then
     verify_scans_support_tools
@@ -169,8 +173,33 @@ execute_second_build() {
 }
 
 fetch_build_cache_metrics() {
-  read_build_scan_metadata
-  fetch_and_read_build_scan_data build_cache_metrics_only "${build_scan_urls[@]}"
+  if [ "$build_scan_publishing_mode" == "on" ]; then
+    read_build_scan_metadata
+    fetch_and_read_build_scan_data build_cache_metrics_only "${build_scan_urls[@]}"
+  else
+    find_build_scan_dumps
+    read_build_scan_dumps
+  fi
+}
+
+find_build_scan_dumps() {
+  local first_build_dir="${EXP_DIR}/first-build_${project_name}"
+  first_build_scan_dump="$(find "$first_build_dir" -maxdepth 1 -regex '^.*build-scan.*\.scan$')"
+  if [ -z "$first_build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the first build"
+  fi
+
+  local second_build_dir="${EXP_DIR}/second-build_${project_name}"
+  second_build_scan_dump="$(find "$second_build_dir" -maxdepth 1 -regex '^.*build-scan.*\.scan$')"
+  if [ -z "$second_build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the second build"
+  fi
+}
+
+read_build_scan_dumps() {
+  local build_scan_csv
+  build_scan_csv="$(invoke_java "$SCANS_SUPPORT_TOOLS_JAR" extract "$first_build_scan_dump" "$second_build_scan_dump")"
+  parse_build_scan_csv "$build_scan_csv"
 }
 
 # Overrides info.sh#print_performance_metrics
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -58,7 +58,7 @@ main() {
 
 verify_scans_support_tools() {
   if [ ! -f "$SCANS_SUPPORT_TOOLS_JAR" ]; then
-    die "ERROR: scans-support-tools.jar is required when running with --no-build-scan-publishing."
+    die "ERROR: scans-support-tools.jar is required when running with --disable-build-scan-publishing."
   fi
 }
 
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -38,7 +38,7 @@ enable_ge=''
 ge_server=''
 interactive_mode=''
 
-# Used with Build Scan publishing disabled
+# Used when Build Scan publishing is disabled
 first_build_scan_dump=''
 second_build_scan_dump=''
 
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -146,7 +146,7 @@ execute_first_build() {
   local init_scripts_dir
   init_scripts_dir="$(init_scripts_path)"
 
-  info "./gradlew --build-cache --scan -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} clean ${tasks}$(print_extra_args)"
+  print_gradle_command
 
   # shellcheck disable=SC2086  # we want tasks to expand with word splitting in this case
   invoke_gradle \
@@ -163,13 +163,21 @@ execute_second_build() {
   local init_scripts_dir
   init_scripts_dir="$(init_scripts_path)"
 
-  info "./gradlew --build-cache --scan -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} clean ${tasks}$(print_extra_args)"
+  print_gradle_command
 
   # shellcheck disable=SC2086  # we want tasks to expand with word splitting in this case
   invoke_gradle \
      --build-cache \
      --init-script "${init_scripts_dir}/configure-local-build-caching.gradle" \
      clean ${tasks}
+}
+
+print_gradle_command() {
+  if [[ "${build_scan_publishing_mode}" == "on" ]]; then
+    info "./gradlew --build-cache --scan -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} clean ${tasks}$(print_extra_args)"
+  else
+    info "./gradlew --build-cache -Dscan.dump -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} clean ${tasks}$(print_extra_args)"
+  fi
 }
 
 fetch_build_cache_metrics() {
--- a/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
+++ b/components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
@@ -190,7 +190,7 @@ find_build_scan_dumps() {
 find_build_scan_dump() {
   local build_name="$1"
   local build_dir="${EXP_DIR}/$build_name-build_${project_name}"
-  build_scan_dump="$(find "$build_dir" -maxdepth 1 -type f -regex '^.*build-scan-.*-.*-.*-.*\.scan')"
+  build_scan_dump="$(find "$build_dir" -maxdepth 1 -type f -regex '^.*build-scan-.*-.*-.*-.*\.scan' | sort | tail -n 1)"
   if [ -z "$build_scan_dump" ]; then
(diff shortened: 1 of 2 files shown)


components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -158,8 +170,33 @@ execute_second_build() {
...
+    fetch_and_read_build_scan_data build_cache_metrics_only "${build_scan_urls[@]}"
+  else
+    find_build_scan_dumps
+    read_build_scan_dumps
+  fi
+}
+
+find_build_scan_dumps() {
+  local first_build_dir="${EXP_DIR}/first-build_${project_name}"
+  first_build_scan_dump="$(find "$first_build_dir" -maxdepth 1 -regex '^.*build-scan.*\.scan$')"
```
erichaagdev, line 184: I'd like some feedback on this approach. I couldn't really find a better way.

erichaagdev, line 184: Should we handle the case where this finds more than one file by this pattern?

jthurne: +1: I think this approach is fine.

The other way we could do this is with `ls -l | grep PATTERN`, but grep behaves differently on different flavors of Unix (for example, the version of grep MacOS ships with is less capable than the one that ships with most distributions of Linux).  In contrast, the `find` command's behavior is more consistent across platforms.

-1: I do suggest adding `-type f` before `-regex` to the find command though. That will limit find to only looking at files and not directories as well.  

It may also be possible to write a more specific regex based on the pattern `build-scan-%s-%s-%d-%s.scan` Something like `^.*build-scan-.*-.*-.*-.*\.scan`, which will further reduce the chances that we pick up a random file in the user's repo.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -19,6 +19,7 @@ readonly SCRIPT_NAME
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo .)")"; pwd)"
 readonly SCRIPT_DIR
 readonly LIB_DIR="${SCRIPT_DIR}/lib"
+readonly SCANS_SUPPORT_TOOLS_JAR="${SCRIPT_DIR}/scans-support-tools.jar"
```
jthurne, line 22: 0: When it comes time to add support to other experiments (like Maven exp 2), we will want to move this to a lib script.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -51,6 +57,12 @@ main() {
   exit_with_return_code
 }
 
+verify_scans_support_tools() {
```
jthurne, line 60: 0: When it comes time to add support to other experiments (like Maven exp 2), we will want to move this to a lib script.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -158,8 +170,33 @@ execute_second_build() {
 }
 
 fetch_build_cache_metrics() {
-  read_build_scan_metadata
-  fetch_and_read_build_scan_data build_cache_metrics_only "${build_scan_urls[@]}"
+  if [ "$build_scan_publishing_mode" == "on" ]; then
```
jthurne, line 173: 0: When it comes time to add support to other experiments, we could move this logic into either a new shared function, or we could update the existing read and fetch functions to behave differently if build scan publishing is disabled.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -158,8 +170,33 @@ execute_second_build() {
...
+  local first_build_dir="${EXP_DIR}/first-build_${project_name}"
+  first_build_scan_dump="$(find "$first_build_dir" -maxdepth 1 -regex '^.*build-scan.*\.scan$')"
+  if [ -z "$first_build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the first build"
+  fi
+
+  local second_build_dir="${EXP_DIR}/second-build_${project_name}"
+  second_build_scan_dump="$(find "$second_build_dir" -maxdepth 1 -regex '^.*build-scan.*\.scan$')"
+  if [ -z "$second_build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the second build"
```
jthurne, line 192: +1 for capitalizing Build Scan (which follows our trademark guidelines).

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
jthurne: I think so. If we find more than one, then I think it makes sense to use the most recent one. If we find more than one Build Scan dump, then the most likely reason is because one or more Build Scan dumps were committed to the repo. We want the one we just generated, which should be more recent than any that were committed to the repo.

It looks Build Scan dump filenames are sortable (they include the build tool version, the Gradle Enterprise plugin version, and then a timestamp), so we could achieve this by adding `| sort | tail -n 1` to the command above. However, this won't work correctly if the extra Build Scan dumps were generated by a newer version of Gradle or the Gradle Enterprise plugin. That seems rather unlikely to me though.

We could also try modifying the find command itself to return the files sorted by oldest-newest instead, based on the file's last modified time.

erichaagdev: I just thought of another option. We could compare which files existed before and after the experiment which match the pattern. We know we only call the build script a single time so there should only be a single file. The only edge case I can think of is if the build itself creates a file which matches that pattern. 🤷

What do you think? I could also just fail for now if we detect more than one file and handle this in a separate issue since it's an edge case.

erichaagdev: I've addressed this with 225a2149591a8f53fd064e489fa11874e128c2de. The simplest approach that should catch the majority of cases is `| sort | tail -n 1`. I tested this by running in interactive mode and generating several Build Scan dumps between the `git clone` and the actual build. Despite there being existing Build Scan dumps, this method chose the most recent.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -158,8 +173,34 @@ execute_second_build() {
...
+}
+
+find_build_scan_dump() {
+  local build_name="$1"
+  local build_dir="${EXP_DIR}/$build_name-build_${project_name}"
+  build_scan_dump="$(find "$build_dir" -maxdepth 1 -type f -regex '^.*build-scan-.*-.*-.*-.*\.scan' | sort | tail -n 1)"
+  if [ -z "$build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the $build_name build"
+  fi
+  eval "${build_name}_build_scan_dump=\${build_scan_dump}"
```
jthurne, line 197: -1: If we wanted to avoid the eval here, we could create a `build_scan_dumps` array instead. This is what we do for normal Build Scans.

You could declare the array like this:

```
build_scan_dumps=()
```

And then you could add to the array here like this:

```
build_scan_dumps+="${build_scan_dump}"
```

To refer to one of the build scan dumps when invoking the Java utility:

```
"${build_scan_dumps[0]" "${build_scan_dumps[1]}"
```

erichaagdev: Great suggestion. Addressed by 5f81c814a56131e1ab6f418c27f4fe7660aaf8b3.

jthurne (approved):
LGTM

sorin-florea (approved):

jthurne (approved):

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
etiennestuder, line 22: Please rename to `build-scan-support-tool.jar`. And also adjust the name in the error message when the jar is not there.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -47,6 +55,12 @@ main() {
   exit_with_return_code
 }
 
+verify_scans_support_tools() {
+  if [ ! -f "$SCANS_SUPPORT_TOOLS_JAR" ]; then
+    die "ERROR: scans-support-tools.jar is required when running with --disable-build-scan-publishing."
```
etiennestuder, line 60: Adjust jar file name as mentioned in the other comment.

etiennestuder: Please change to `ERROR: build-scan-support-tool.jar is required when using --disable-build-scan-publishing.`

components/scripts/lib/gradle.sh
```diff
@@ -26,8 +26,13 @@ invoke_gradle() {
     args+=("-Pcom.gradle.enterprise.build_validation.server=${ge_server}")
   fi
 
+  if [ "$build_scan_publishing_mode" == "off" ]; then
```
etiennestuder, line 29: I would invert this if-statement and handle `on` case first. Just like you do further up in `print_gradle_command()`.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -47,6 +55,12 @@ main() {
   exit_with_return_code
 }
 
+verify_scans_support_tools() {
```
etiennestuder, line 58: I suggest to rename to `verify_build_scan_support_tool_exists`.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -157,9 +171,45 @@ execute_second_build() {
...
+}
+
+find_build_scan_dumps() {
+  find_build_scan_dump "first"
+  find_build_scan_dump "second"
+}
+
+find_build_scan_dump() {
+  local build_name="$1"
+  local build_dir="${EXP_DIR}/$build_name-build_${project_name}"
```
etiennestuder, line 199: Does this start deep enough in the folder hierarchy to not find a scan dump from a previous experiment run?

etiennestuder: I think so, but just wanted to ask for confirmation.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
```diff
@@ -157,9 +171,45 @@ execute_second_build() {
...
+  build_scan_dump="$(find "$build_dir" -maxdepth 1 -type f -regex '^.*build-scan-.*-.*-.*-.*\.scan' | sort | tail -n 1)"
+  if [ -z "$build_scan_dump" ]; then
+    die "ERROR: No Build Scan dump found for the $build_name build"
+  fi
+  build_scan_dumps+=("${build_scan_dump}")
```
etiennestuder, line 204: Is this conciously not in an `else` block?

etiennestuder (changes requested):

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Yes, it does. This is the exact path to the root project directory (e.g. `.data/02-validate-local-build-caching-same-location/20221109T135440-636c0580/first-build_java-ordered-properties`).

However, now I'm wondering how this should work when `-p` / `--project-dir` is used? 

cc @jthurne

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Yes, since we call `die` on line 202 when the `if` evaluates to `true`. 

Should I put it in an `else`?

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Changed in 959de12707c38c55194c27955284d8a8bc5830ca

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Changed in 959de12707c38c55194c27955284d8a8bc5830ca

components/scripts/lib/gradle.sh
erichaagdev: Changed in 17fe64ac094a951f47340572dc60051a9419a40d

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
erichaagdev: Changed in 959de12707c38c55194c27955284d8a8bc5830ca

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
etiennestuder: -p should only influence the dir from which the project is invoked. I think all is fine. But, @jthurne please weigh in here.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
etiennestuder: Ah, makes sense. Resolving.

components/scripts/gradle/03-validate-local-build-caching-different-locations.sh
jthurne: If the user provides `-p/--project-dir, before invoking the build tool, the scripts will cd to the directory given by `-p` within the repo, and then invoke Gradle or Maven.  This means that any output generated by Gradle will also be in the subdirectory, including the build scan dump.   As such, I do think there may be a problem here.

There are two ways we could correct this:

  1. If `project_dir` is set, include it in the path where we look for the build scan dump.
  2. remove the `-maxdepth` argument from the find command so that it finds build scan dumps that may be in a subdirectory.
 
The first option is most likely the "safest".

erichaagdev: Thanks for the input. I will address this in a different PR. I have created #217 for this.

PR #185, opened 2022-10-31
Improve text and guidance when Build Scan publishing is disabled

### LINKED ISSUE #183: Texts and guidance shown to the user is improved when no build scans are published
- remove `Running experiment with Build Scan publishing disabled.` from the debug mode

```
main() {
  if [[ "$build_scan_publishing_mode" == "off" ]]; then
    debug "Running experiment with Build Scan publishing disabled."
  fi
```

- when using `-h` option for exp3, include it in the list of options (see short/long flag, description, and option ordering):

```
...
-a, --args                         Specifies additional arguments to pass to Gradle.
-s, --gradle-enterprise-server     Specifies the URL for the Gradle Enterprise server to connect to.
-e, --enable-gradle-enterprise     Enables Gradle Enterprise on a project not already connected.
-x, --no-build-scan-publishing     Disables the publication of build scans on the invoked builds.
-f, --fail-if-not-fully-cacheable  Terminates with exit code 3 if the build is not fully cacheable.
...
```

- use the short form (`-x`) in the _Command Line Invocation_ section when running in interactive mode (and order as in the example below, i.e. before -f):

```
Command Line Invocation
-----------------------
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties -b ge -t build -s https://ge.solutions-team.gradle.com/ -x -f
```
- adjust the text shown in the summary to the following:

```
Build scan first build:     <publication disabled>
Build scan second build:    <publication disabled>
```

- Remove the entire section `Preparation II. - Ensure Gradle Enterprise API access` from the interactive mode when build scan publication is disabled (we can live with the non-perfection of _Preparation I._ being shown for the prior preparation step)

```
Preparation II. - Ensure Gradle Enterprise API access`  ...
```

- Replace the text below, shown to the user in the section `Measure build` during the interactive mode when build scan publication is disabled (leave all other text in that section as it currently is for both cases)

```
Some of the build scan data will be fetched from the build scans produced by the
two builds to assist you in your investigation.

Fetching build scan data, done.
```
becomes 
```
Some of the build scan data will be extracted from the locally stored, intermediate
build data produced by the two builds to assist you in your investigation.

Extracting build scan data, done.
```

- Add an extra sentence, shown to the user in the section `Measure build` during the interactive mode when build scan publication is disabled (leave all other text in that section as is for both cases) 

```
The ‘Summary’ section below captures the configuration of the experiment.
```
becomes
```
The ‘Summary’ section below captures the configuration of the experiment. No
build scans are available for inspection since publishing was disabled for the
experiment.
```

(see original [comment](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/178#issuecomment-1296256047))

jthurne:
### ✅  remove Running experiment with Build Scan publishing disabled. from the debug mode

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing \
  -- debug
```

image

### ✅  when using -h option for exp3, include it in the list of options (see short/long flag, description, and option ordering):

```
./03-validate-local-build-caching-different-locations.sh --help
```

image

### ✅ use the short form (-x) in the Command Line Invocation section when running in interactive mode (and order as in the example below, i.e. before -f):

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing \
  --interactive
```

image

### ✅ adjust the text shown in the summary to the following:

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing
```

image

### ✅  Remove the entire section Preparation II. - Ensure Gradle Enterprise API access from the interactive mode when build scan publication is disabled (we can live with the non-perfection of Preparation I. being shown for the prior preparation step)

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing \
  --interactive
```

image

### ✅  Replace the text below, shown to the user in the section Measure build during the interactive mode when build scan publication is disabled (leave all other text in that section as it currently is for both cases)...

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing \
  --interactive
```

image

**NOTE**: This PR does not update the message returned when extracting data from the build scan dumps. That will need to be done as part of #179.

### ✅ Add an extra sentence, shown to the user in the section Measure build during the interactive mode when build scan publication is disabled (leave all other text in that section as is for both cases)...

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/etiennestuder/java-ordered-properties \
  -b ge \
  -t assemble \
  --no-build-scan-publishing \
  --interactive
```

image

Closes #183

### DIFF
--- a/components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
+++ b/components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2034  # It is common for variables in this auto-generated file to go unused
 # Created by argbash-init v2.10.0
-# ARG_OPTIONAL_BOOLEAN([build-scan-publishing],[],[],[on])
+# ARG_OPTIONAL_BOOLEAN([build-scan-publishing],[x],[],[on])
 # ARG_OPTIONAL_BOOLEAN([fail-if-not-fully-cacheable],[f],[])
 # ARG_HELP([This function is overridden later on.])
 # ARG_VERSION([print_version],[v],[version],[])
@@ -28,6 +28,7 @@ function print_help() {
   print_option_usage -a
   print_option_usage -s
   print_option_usage -e
+  print_option_usage -x
   print_option_usage -f
   print_option_usage -v
   print_option_usage -h
--- a/components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
+++ b/components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2034  # It is common for variables in this auto-generated file to go unused
 # Created by argbash-init v2.10.0
-# ARG_OPTIONAL_BOOLEAN([build-scan-publishing],[x],[],[on])
+# ARG_OPTIONAL_BOOLEAN([build-scan-publishing],[],[],[on])
+# ARG_OPTIONAL_BOOLEAN([disable-build-scan-publishing],[x])
 # ARG_OPTIONAL_BOOLEAN([fail-if-not-fully-cacheable],[f],[])
 # ARG_HELP([This function is overridden later on.])
 # ARG_VERSION([print_version],[v],[version],[])
--- a/components/scripts/lib/config.sh
+++ b/components/scripts/lib/config.sh
@@ -271,7 +271,7 @@ print_command_to_repeat_experiment() {
   fi
 
   if [[ "${build_scan_publishing_mode}" == "off" ]]; then
-    cmd+=("--no-build-scan-publishing")
+    cmd+=("-x")
   fi
 
   if [[ "${fail_if_not_fully_cacheable}" == "on" ]]; then
--- a/components/scripts/lib/config.sh
+++ b/components/scripts/lib/config.sh
@@ -60,6 +60,11 @@ process_arguments() {
   else
     build_scan_publishing_mode=on
   fi
+  if [ -n "${_arg_disable_build_scan_publishing+x}" ]; then
+    if [[ "${_arg_disable_build_scan_publishing}" == "on" ]]; then
+      build_scan_publishing_mode=off
+    fi
+  fi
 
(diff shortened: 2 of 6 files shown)


components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
```diff
@@ -2,6 +2,7 @@
 # shellcheck disable=SC2034  # It is common for variables in this auto-generated file to go unused
 # Created by argbash-init v2.10.0
 # ARG_OPTIONAL_BOOLEAN([build-scan-publishing],[],[],[on])
+# ARG_OPTIONAL_BOOLEAN([disable-build-scan-publishing],[x])
```
jthurne, line 5: This is a workaround for a limitation with how Argbash handles boolean CLI switches. Basically, there is no way to set a shorthand for a boolean CLI switch to set the its value to "off".

The actual --disable-build-scan-publishing CLI argument won't be presented to users, but if a user happens to use it, it will do the right thing. There is some potentially odd behavior if both flags are used at the same time, but this seems unlikely (since users won't know there are multiple flags unless they inspect the script source code).

components/scripts/lib/config.sh
```diff
@@ -60,6 +60,11 @@ process_arguments() {
   else
     build_scan_publishing_mode=on
   fi
+  if [ -n "${_arg_disable_build_scan_publishing+x}" ]; then
```
jthurne, line 63: This is part of the workaround I mentioned earlier.

components/scripts/lib/cli-parsers/gradle/03-cli-parser.m4
etiennestuder: Thanks for the explanation.

etiennestuder (approved):
Thanks, @jthurne . Fine-tuning communicated via Slack.

PR #200, opened 2022-11-07
Improve fetch-build-scan-data-cmdline-tool startup perf by setting Picocli sys prop

jthurne:
Closure support is only useful when using Picocli with Groovy. Disabling
support for closures reportedly improves startup performance. See
[the Picocli 4.7.0 release notes](https://github.com/remkop/picocli/blob/main/RELEASE-NOTES.md#-picocli-470).

### DIFF
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/Main.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/Main.java
@@ -7,6 +7,7 @@ import picocli.CommandLine.Help.ColorScheme;
 
 public class Main {
     public static void main(String[] args) {
+        System.setProperty("picocli.disable.closures", "true");  // improves startup time
         ColorScheme colorScheme = CommandLine.Help.defaultColorScheme(CommandLine.Help.Ansi.AUTO);
         CommandLine cmdLine = new CommandLine(new FetchBuildValidationDataCommand(colorScheme))
             .setExecutionExceptionHandler(new PrintExceptionHandler())


erichaagdev:
I tried to test this by running 100 invocations of `--help` and comparing the before and after results:

image

Each row is the total time in seconds it took to run 100 invocations of `--help`.

There seems to be no noticeable difference, but I guess it doesn't hurt to merge this anyway.

erichaagdev (approved):

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/Main.java
```diff
@@ -7,6 +7,7 @@
 
 public class Main {
     public static void main(String[] args) {
+        System.setProperty("picocli.disable.closures", "true");  // improves startup time
```
etiennestuder, line 10: Typically, sys props are set by the invoker, IMO. Are we certain that the sys prop set here is actually picked up, and not set too late? If so, please merge the PR to main.

jthurne: I went this route so as to avoid having to expose this detail to the code that invokes the script, but it's a good point that maybe it is ineffective here.

@erichaagdev would you be able to test an alternative implementation where the system property is set when the utility is invoked by the scripts (using `-D`)?  Maybe that's why we aren't seeing any kind of performance gain with this change.

erichaagdev:
We have decided to close this PR as there seems to be no noticeable performance gains by setting the `picocli.disable.closures` system property to `true`.

PR #210, opened 2022-11-16
Update README to use version 2.1 of validation scripts

runningcode:


erichaagdev (approved):
👍

runningcode:
Let's update this to use the new tag that @jprinet will create before merging.

PR #214, opened 2022-11-18
Adjust the cross-platform-tests target operating systems 

jthurne:
Some operating system versions have been deprecated, while others have been added.

Note that we can't test against windows-2022 because WSL v1 is not supported on Windows Server 2022.


etiennestuder (approved):
thanks, @jthurne

PR #219, opened 2022-11-23
Add fail if not fully cacheable flag to GH composite workflows

jprinet:
I changed the processing of `enableGradleEnterprise` flag as well to allow true/false configuration.
Previous behavior was to enable the flag as soon as the value was not empty

### DIFF
--- a/.github/actions/gradle/experiment-2/action.yml
+++ b/.github/actions/gradle/experiment-2/action.yml
@@ -26,6 +26,9 @@ inputs:
   enableGradleEnterprise:
     description: "Enables Gradle Enterprise on a project not already connected"
     required: false
+  failIfNotFullyCacheable:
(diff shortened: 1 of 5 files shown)


.github/actions/gradle/experiment-2/action.yml
```diff
@@ -26,6 +26,9 @@ inputs:
   enableGradleEnterprise:
     description: "Enables Gradle Enterprise on a project not already connected"
     required: false
+  failIfNotFullyCacheable:
+    description: "Fail the build if it is not fully cacheable"
```
etiennestuder, line 30: I would use the text `Terminates with exit code 3 if the build is not fully cacheable` (same as on the cmd line help text) and apply to all 5 experiments in this PR.

etiennestuder (changes requested):

PR #221, opened 2022-11-24
GH composite actions with enriched summary

jprinet:
This would allow to have the full experiment receipt directly accessible (with clickable links)


etiennestuder:
@jprinet Can you please paste a link to an example GH build that shows the new behavior? Thanks.

.github/actions/gradle/experiment-1/action.yml
```diff
@@ -103,8 +103,8 @@ runs:
         
         echo "buildScanFirstBuild=$BUILD_SCAN_1" >> $GITHUB_OUTPUT
         echo "buildScanSecondBuild=$BUILD_SCAN_2" >> $GITHUB_OUTPUT
```
etiennestuder, line 105: I would remove this two statements, now that we have the rich summary:
image

etiennestuder, line 105: I would continue to list the receipt files under the `Artifacts` section: (in case someone wants to download them)

image

etiennestuder (changes requested):

.github/actions/gradle/experiment-1/action.yml
etiennestuder: It seems this is still the case. Please resolve the comment if it is obsolete.

etiennestuder: Hm, I catually do not see it in the output anymore, but I wonder why not, given the statements are still in the yml file.

jprinet: Those are actually not printed but actual outputs of the composite step.
I would still keep them as it might be useful for automation imo (ie. subsequent calls to the API).

Please let me know your preference.

jprinet: Yes I kept the receipts as workflow artifacts

etiennestuder: Ah. I see. Let's leave it then as it is on this PR.

etiennestuder (approved):
LGTM

PR #229, opened 2022-12-06
Users understand the cmd line arguments as soon as they have provided all information in interactive mode

jthurne:
If the experiment needs to be interrupted for some reason, it's rather annoying
to have to re-enter all of the settings when re-running the experiment. We show
the command to repeat the experiment at the end of the summary, but that's only
helpful if the script completes normally.

This commit updates interactive mode to also show the command to repeat the
experiment immediately after all of the configuration options have been
collected, but before executing any of the builds.

### DIFF
--- a/components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
+++ b/components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
@@ -92,6 +92,7 @@ wizard_execute() {
   explain_collect_mapping_file
   print_bl
   collect_mapping_file
+  explain_command_to_repeat_experiment_after_collecting_parameters
 
   print_bl
   parse_build_scan_urls
@@ -311,8 +312,8 @@ EOF
   wait_for_enter
 }
 
-#Overrides config.sh#print_command_to_repeat_experiment
-print_command_to_repeat_experiment() {
+#Overrides config.sh#generate_command_to_repeat_experiment
+generate_command_to_repeat_experiment() {
   local cmd
   cmd=("./${SCRIPT_NAME}")
   cmd+=("-1" "${build_scan_urls[0]}")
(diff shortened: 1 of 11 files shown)


runningcode:
This is a very valuable change!

components/scripts/gradle/04-validate-remote-build-caching-ci-ci.sh
```diff
@@ -311,8 +312,8 @@ EOF
   wait_for_enter
 }
 
-#Overrides config.sh#print_command_to_repeat_experiment
-print_command_to_repeat_experiment() {
+#Overrides config.sh#generate_command_to_repeat_experiment
+generate_command_to_repeat_experiment() {
```
jthurne, line 316: I thought about whether or not we could get rid of this function override (ant the override in the equivalent Maven experiment script), but the issue is that this script does set some of the variables that would result in invalid flags being set for this particular script.

I suppose we could add special handling for this script specifically (test if the experiment is Gradle exp 04 and skip evaluating the invalid flags), but that has its own downsides.

erichaagdev: TBH, I think the override is fine. This particular function is overridden in `exp4-gradle`, `exp5-gradle`, `exp3-maven`, and `exp4-maven`. It would take a lot of special handling to make a single function handle all cases.

erichaagdev:
I ran tests for each experiment but did not document each outcome for the sake of brevity. My test steps were as follows:

1. Run the experiment in interactive mode
2. Copy the repeat command from the new location
3. Compare the copied repeat command with the summary's repeat command
4. Re-run the experiment using the copied repeat command

erichaagdev (approved):
👍

etiennestuder (approved):
LGTM

PR #239, opened 2022-12-12
Force CaptureBuildScansListener to read GE server URL after it is set

### LINKED ISSUE #234: Enforce Gradle Enterprise URL is set when using it in the Capture Build Scan URL Maven extension
## Context
When running a Maven build with the automated validation scripts, a build scan link is added to filter the build scans related to a given run.

This is done through the [Capture Build Scan URL Maven extension](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/6ba97dad760344b0df6f37a5355149a07cb968a1/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java#L54).

## Issue
We can not necessarily ensure the ordering in which extensions are applied and that can lead to consuming the Gradle Enterprise server URL before it is actually set by the Gradle Enterprise main extension, which in turn leads to a NullPointerException [there](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/6ba97dad760344b0df6f37a5355149a07cb968a1/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java#L61).

## Proposal
Encapsulate the access in a `api.buildFinished` block to ensure the value is set.

jprinet:
Fix #234

### DIFF
--- a/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
+++ b/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
@@ -40,7 +40,7 @@ public class CaptureBuildScansListener implements GradleEnterpriseListener {
     }
 
     private static void addCustomDataOnBuildFinished(BuildScanApi buildScan) {
-        // Encapsulation in buildFinished block to ensure GE server URL is set (required by buildScan.getServer())
+        // Links are set in buildFinished block to ensure buildScan.getServer() is set (required by addCustomValueAndSearchLink())
         buildScan.buildFinished(ignored -> {
             String expId = System.getProperty("com.gradle.enterprise.build_validation.expId");
             addCustomValueAndSearchLink(buildScan, "Experiment id", expId);
--- a/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
+++ b/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
@@ -80,7 +80,7 @@ public class CaptureBuildScansListener implements GradleEnterpriseListener {
             try (FileWriter fw = new FileWriter(EXPERIMENT_DIR + "/build-scans.csv", true);
                 BufferedWriter bw = new BufferedWriter(fw);
                 PrintWriter out = new PrintWriter(bw)) {
-                out.println(String.format("%s,%s,%s,%s", rootProject.getName(), baseUrl, scan.getBuildScanUri(), scan.getBuildScanId()));
+               out.println(String.format("%s,%s,%s,%s", rootProject.getName(), baseUrl, scan.getBuildScanUri(), scan.getBuildScanId()));
             } catch (IOException e) {
                 logger.error("Unable to save scan data to build-scans.csv: " + e.getMessage(), e);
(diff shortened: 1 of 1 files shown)


jprinet:
cc @facewindu

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
```diff
@@ -76,8 +78,8 @@ private void capturePublishedBuildScan(BuildScanApi buildScan, MavenProject root
             String baseUrl = String.format("%s://%s%s", scan.getBuildScanUri().getScheme(), scan.getBuildScanUri().getHost(), port);
 
             try (FileWriter fw = new FileWriter(EXPERIMENT_DIR + "/build-scans.csv", true);
-                 BufferedWriter bw = new BufferedWriter(fw);
-                 PrintWriter out = new PrintWriter(bw)) {
+                BufferedWriter bw = new BufferedWriter(fw);
+                PrintWriter out = new PrintWriter(bw)) {
```
runningcode, line 82: not related to this PR: my understanding is that we need to close this `Printwriter` which is the outermost writer. Closing the outermost writer will take care of closing the inside ones - in this case `bw` and `fw`.

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
```diff
@@ -36,17 +35,20 @@ public void configure(GradleEnterpriseApi api, MavenSession session) throws Exce
...
+        addCustomDataOnBuildFinished(buildScan);
         capturePublishedBuildScan(buildScan, rootProjectExtractor.extractRootProject(session));
     }
 
-    private static void addCustomData(BuildScanApi buildScan) {
-        String expId = System.getProperty("com.gradle.enterprise.build_validation.expId");
-        addCustomValueAndSearchLink(buildScan, "Experiment id", expId);
-        buildScan.tag(expId);
+    private static void addCustomDataOnBuildFinished(BuildScanApi buildScan) {
+        // Encapsulation in buildFinished block to ensure GE server URL is set (required by buildScan.getServer())
```
runningcode, line 43:
```suggestion
        // Links are set in buildFinished block to ensure buildScan.getServer() is set (required by addCustomValueAndSearchLink())
```

runningcode (approved):
Just some nits from my side but LGTM. Please wait for the others to review this as well.

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
jprinet: Your comment made me realize that the previous indentation was there on purpose. I'll revert my change.

The current implementation is fine as the writers are part of the `try-with-resources` definition and will all be closed in reverse order of definition.
Closing an already closed stream is a `no-op`.

runningcode: ah i see that now, i missed that they were all inside the try-with-resources :D

jthurne (approved):

etiennestuder (approved):

PR #241, opened 2022-12-12
Do not use pushd "" as behavior can be unexpected when used with CDPATH

### LINKED ISSUE #236: Scripts work when users have configured pushd directory
More details to be provided by @jprinet soon.

jprinet:
fix #236

### DIFF
--- a/components/scripts/lib/paths.sh
+++ b/components/scripts/lib/paths.sh
@@ -46,7 +46,7 @@ relative_path() {
     printf '%s\n' "$result"
 }
 
-pushd_with_project_dir() {
+pushd_to_project_dir() {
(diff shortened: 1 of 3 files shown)


components/scripts/lib/paths.sh
```diff
@@ -46,9 +46,19 @@ relative_path() {
     printf '%s\n' "$result"
 }
 
+pushd_with_project_dir() {
```
jthurne, line 49: I suggest renaming this to `pushd_to_project_dir` since we're changing directories to the project_dir (it's similar to something like `cd_to_project_dir`).

jthurne (approved):
LGTM, just a small function naming suggestion.

etiennestuder (approved):

PR #263, opened 2022-12-29
Always read project name from Build Scan data

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. ⛄ #263
4. #264
5. #265
6. #266
7. #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

ℹ️ This PR continues the implementation of https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/206. This felt like a good place to stop. More PRs will be coming.

This PR introduces the following changes:

- Removed all reading and writing of project name to the `build-scans.csv` file
- Removed many redundant occurrences of reading the `build-scans.csv` file
  - In most cases it was being read twice, or three times when in interactive mode
- Added debug logging for `build-scans.csv` that was very useful when testing this PR

These changes do not introduce any visual changes in user facing output.  

## Testing

I ran all of the following experiments and inspected the output. I ensured that:

- All summary information was displayed as expected, including the project name
- The `build-scans.csv` file was only ever being read once

```sh
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' --debug
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -i --debug
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x --debug
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -i -x --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -i --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -i -x --debug
[...]

PR #262, opened 2022-12-28
Include performance characteristics and support `-x` for Gradle experiment 1 

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. ⛄ #262
3. #263
4. #264
5. #265
6. #266
7. #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

ℹ️ This PR continues the implementation of https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/206. This felt like a good place to stop. More PRs will be coming.

This PR introduces the following changes:

- The summary of Gradle experiment 1 now includes a `Performance Characteristics` section that includes `Effective task execution duration`
- Gradle experiment 1 can now be run with `-x` / `--disable-build-scan-publishing`

#### Non-interactive

```sh
./01-validate-incremental-building.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build'
```

image

#### Interactive

```sh
./01-validate-incremental-building.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build' \
  -i
```

image

image

image

### Disabled Build Scan publishing

⚠️ The _Performance Characteristics_ section is intentionally empty with `-x` because the `build-scan-support-tool.jar` does not return the required `Effective task execution duration` field. Once this required field is returned, the section will include the new `Realized build time savings` field as expected.

#### Non-interactive

```sh
./01-validate-incremental-building.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build' \
  -x
```

image

#### Interactive

```sh
./01-validate-incremental-building.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build' \
  -i \
  -x
```

image

image

image

PR #261, opened 2022-12-28
Include realized build time savings in experiment summary

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. ⛄ #261
2. #262
3. #263
4. #264
5. #265 
6. #266
7. #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

ℹ️ This PR only partially implements #206. This felt like a good place to stop. More PRs will be coming.

This PR introduces the following changes:

- The API command line tool now includes `Effective task execution duration` in the CSV response
- The _Build Caching Leverage_ section of the experiment summary has been renamed to _Performance Characteristics_
- `Realized build time savings` is now included in the _Performance Characteristics_ section of the experiment summary

### Example

```
Performance Characteristics
----------------------
Realized build time savings:  1m 7.377s wall-clock time (from 1m 11.542s to 4.165s)
Avoided cacheable tasks:      292 tasks, 4m 44.680s total saved execution time
Executed cacheable tasks:       1 tasks, 0.045s total execution time
Executed non-cacheable tasks: 292 tasks, 10.686s total execution time

WARNING: Not all cacheable tasks' outputs were taken from the build cache in the second build. This reduces the savings in task execution time.
```

### Notes

- I ensured this change was backwards compatible with the current version of the `build-scan-support-tool.jar`. If the new `Effective task execution duration` field is not included in the command line CSV response then `Realized build time savings` will not be displayed.
- For the display format of `Realized build time savings` I decided to take @runningcode's [suggestion](https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/206#issuecomment-1366014004). I even tried a few display variations myself and liked his suggestion the best. Let me know if the format should be changed.

#### Experiment 1

```sh
./01-validate-incremental-building.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build'
```

_No visual differences._

image

##### Online

```sh
./02-validate-local-build-caching-same-location.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build'
```

image

##### Offline

```sh
./02-validate-local-build-caching-same-location.sh \
  -r git@github.com:gradle/ge-solutions.git \
  -p sample-projects/gradle/7.x/ge \
  -t 'build' \
  -x
```

_No visual differences._

image

##### Online

```sh
./03-validate-local-build-caching-different-locations.sh \
[...]

PR #121, opened 2022-07-26
Gradle init script improvements

bigdaz:
Various refactorings and improvements to the init-scripts used to instrument Gradle:
- Combine 3 separate init-scripts that were configuring GE into a single script.
- Improve error message when build-validation was run on a non-GE project (without `-e`).
- Avoid duplicate plugin application when running on a GE-enabled project with `-e`.

As well as improving init-script behaviour, these increase the similarity between init-scripts in this project and the TC Build Scan plugin.

PR #264, opened 2022-12-29
Gradle init script improvements

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. ⛄ #264
5. #265
6. #266
7. #267
8. #272
9. #274 
10. #283

⛄ = you are here

## Summary

This PR is a rebased version of #121 from @bigdaz. I decided to create a new branch and PR in order to preserve his branch as it. That way I could easily switch between the two as I verified these changes. Once this PR is merged #121 can be closed.

|   |
| --- |
| ⚠️ While testing these changes, I discovered a bug that exists in the #121 version when running with Gradle 5.x. When running the scripts with `-e` both `com.gradle.build-scan` and `com.gradle.enterprise` are applied. This can be seen in this [Build Scan](https://scans.gradle.com/s/a7dz6nhqskzne/build-dependencies?toggled=W1swXSxbMCwwXSxbMV0sWzEsMF1d). It seems this scenario [causes an error when applying CCUD](https://scans.gradle.com/s/a7dz6nhqskzne/failure?focused-stack-frame=0-1-2-3#1).<br><br>I was able to resolve this by removing `--scan` as a parameter to the build validation scripts' Gradle invocation. Since the initialization scripts take care of configuring and setting up Gradle Enterprise I don't think the `--scan` parameter is required.<br><br>However, `--scan` is still displayed in the printed Gradle invocation command. This is somewhat misleading now. In my opinion, it was  also misleading before since the displayed command does not include the initialization scripts that perform additional configuration transparent to the user.  |

The only other change from #121 is that I removed all `logger` calls in the initialization scripts since they created unnecessary noise in the validation scripts output that was not present before. 

## Testing

In order to test this I ran the following experiments and verified a successful outcome. 

The following parameter combinations were tested with Gradle `5.x`, `6.x`, and `7.x`:

- online
- offline (`-x`)
- online, enable GE (`-e`)
- offline (`-x`), enable GE  (`-e`)

```sh
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/5.x/ge -t 'build'
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/5.x/ge -t 'build' -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/5.x/no-ge -t 'build' -e -s https://ge.solutions-team.gradle.com
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/5.x/no-ge -t 'build' -e -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/6.x/ge -t 'build'
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/6.x/ge -t 'build' -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/6.x/no-ge -t 'build' -e -s https://ge.solutions-team.gradle.com
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/6.x/no-ge -t 'build' -e -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build'
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/no-ge -t 'build' -e -s https://ge.solutions-team.gradle.com
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/no-ge -t 'build' -e -x
./04-validate-remote-build-caching-ci-ci.sh -1 https://ge.solutions-team.gradle.com/s/fvwne3ht4xgjg -2 https://ge.solutions-team.gradle.com/s/tjiwbi2b7oioo
./05-validate-remote-build-caching-ci-local.sh -1 https://ge.solutions-team.gradle.com/s/lc73bw56kdkxi
```


jthurne:
> However, --scan is still displayed in the printed Gradle invocation command. This is somewhat misleading now. In my opinion, it was also misleading before since the displayed command does not include the initialization scripts that perform additional configuration transparent to the user.

The intent with the displayed command is to show the user a simple command they can use themselves when running Gradle directly to achieve a similar result. With that in mind, we don't print out the full command used so as not to distract or misslead the user.

These initialization scripts have become sufficiently complex that I think many users would have difficulty interpreting them or understanding what they do.  Which is fine, but it's a reason why we don't necessarily want to display their use in the command that we execute.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -1,69 +1,160 @@
 import org.gradle.util.GradleVersion
 import java.nio.charset.StandardCharsets;
 
+if (GradleVersion.current() < GradleVersion.version("5.0")) {
+    throw new IllegalStateException("Build validation not supported for Gradle ${GradleVersion.current()}. Upgrade your project's build to Gradle 5 or newer.")
```
jthurne, line 5:
```suggestion
    throw new IllegalStateException("The Gradle Enterprise Build Validation Scripts are not supported with Gradle ${GradleVersion.current()}. Please upgrade your project's build to Gradle 5 or newer.")
```

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -1,69 +1,160 @@
...
+        }
+        def errorFile = new File(experimentDir, "build-scan-publish-error.txt")
+        buildScan.onError { error ->
+            errorFile.text = error
+        }
+    }
+}
 
-    clearWarnings()
+// register buildScanPublished listener and optionally apply the GE / Build Scan plugin
```
jthurne, line 75: -1: I don't see a buildScanPublished listener being registered below anywhere.

```suggestion
// Optionally apply the GE / Build Scan plugin
```

components/scripts/lib/gradle.sh
```diff
@@ -10,20 +10,17 @@ invoke_gradle() {
     cd "${project_dir}" > /dev/null 2>&1 || die "ERROR: The subdirectory ${project_dir} (set with --project-dir) does not exist in ${project_name}." "${INVALID_INPUT}"
   fi
 
+  args+=(--init-script "${INIT_SCRIPTS_DIR}/configure-gradle-enterprise.gradle")
+
   if [ "$enable_ge" == "on" ]; then
-    args+=(--init-script "${INIT_SCRIPTS_DIR}/enable-gradle-enterprise.gradle")
+    args+=("-Dcom.gradle.enterprise.build_validation.enableGradleEnterprise=true")
```
jthurne, line 16: 0: I'm curious, why are we setting these as system properties instead of as project properties?

jthurne:
> Once this PR is merged #121 can be closed.

Would it be okay if we closed it now?  I've gotten confused a few times between #121 and this PR.  Closing a PR doesn't delete the underlying branch, so you could still compare your branch against Daz's original branch.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
erichaagdev: The listener is being applied [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/a2e2bf985267d51e5b2c59502c72bba7da28e811/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle#L107) and [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/a2e2bf985267d51e5b2c59502c72bba7da28e811/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle#L139).

components/scripts/lib/gradle.sh
erichaagdev: That's a great question. I am not sure. I only continued the pattern started by @bigdaz in #121. 

Can you think of any reason not to use a project property?

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
jthurne: So it is! 👍

erichaagdev:
A positive side effect of removing `--scan` is that if a build is not actually configured with Gradle Enterprise, then it will not try to publish a Build Scan to scans.gradle.com. For example:

```
./03-validate-local-build-caching-different-locations.sh -r https://github.com/android/nowinandroid -t assembleDemoDebug
```

If we run the above without these changes, the build will fail with the following error:

image

As you can see, it is asking to accept the terms of service so that a Build Scan may be published to scans.gradle.com. This is risky when running the scripts for a project whose Build Scan should not be made publicly available. With these changes, the user will see the following instead:

image

components/scripts/lib/gradle.sh
bigdaz: It was my understanding that project properties aren't available when an init script evaluates. But I could well be mistaken.

erichaagdev: That makes sense @bigdaz. I'm having trouble finding it anywhere in the documentation, but I did make a small sample project to test.

```
* What went wrong:
A problem occurred evaluating initialization script.
> The default project is not yet available for build.
```

image

PR #270, opened 2023-01-03
Add retry mechanism to curl instruction

### LINKED ISSUE #259: Improve error handling of download action
The download action occasionally fails. See the following examples:

- https://github.com/gradle/gradle-enterprise-build-config-samples/actions/runs/3759719719/jobs/6389623243#step:2:28
- https://github.com/gradle/gradle-enterprise-oss-projects/actions/runs/3760975144/jobs/6392238897#step:6:52

It's unclear what is causing this. [Rate limiting](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting) is one possibility though GitHub API rate limiting is pretty generous. 

As a first pass at tackling this issue, I propose we add some error handling around the HTTP call. We should check the response status is what we expect (e.g. 200) and log to the console if it's something else. We could also check the response headers to provide a better error message. In the case of a rate limit, there will be a `x-ratelimit-remaining: 0` header [according to the documentation.](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limit-http-headers).

jprinet:
fixes #259

### DIFF
--- a/.github/actions/gradle/download/action.yml
+++ b/.github/actions/gradle/download/action.yml
@@ -12,24 +12,18 @@ runs:
     - name: Download latest Gradle build validation scripts
       run: |
         # Set curl general options
-        curlOpts="--silent --fail --show-error --retry 3 --retry-delay 5 --location"
+        curlOpts=(--silent --fail --show-error --retry 3 --retry-delay 5 --location)
 
         # Set authorization header if needed
         if [ ! -z "${{ inputs.token }}" ]; then
-          curlOpts="$curlOpts --header 'Authorization: Bearer ${{ inputs.token }}'"
+          curlOpts+=(--header 'Authorization: Bearer ${{ inputs.token }}')
         fi
 
-        # Set curl params to get latest release data of the build validation scripts
-        curlParamsGetLatestReleaseData="$curlOpts https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest"
-
         # Get the download url of the latest release
-        downloadUrl=$(eval curl $curlParamsGetLatestReleaseData | jq -c '.assets[] | select(.content_type == "application/zip")' | jq -r .browser_download_url | grep -v maven)
-
-        # Set curl params to download the build validation scripts
-        curlParamsDownloadScripts="$curlOpts --output gradle-enterprise-gradle-build-validation.zip $downloadUrl"
+        downloadUrl=$(curl "${curlOpts[@]}" https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest | jq -c '.assets[] | select(.content_type == "application/zip")' | jq -r .browser_download_url | grep -v maven)
 
         # Download the latest version of the build validation scripts
-        eval curl $curlParamsDownloadScripts
+        curl "${curlOpts[@]}" --output gradle-enterprise-gradle-build-validation.zip $downloadUrl
 
         # Unzip the downloaded build validation scripts
         unzip -q -o gradle-enterprise-gradle-build-validation.zip
--- a/.github/actions/gradle/download/action.yml
+++ b/.github/actions/gradle/download/action.yml
@@ -11,22 +11,25 @@ runs:
   steps:
     - name: Download latest Gradle build validation scripts
       run: |
-        authHeader=""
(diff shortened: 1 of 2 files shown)


jprinet:
This won't fix the intermittent issues when due to rate limiting, will make it even worse in fact, but will help diagnosing better.
If rate limiting is proven to be the root cause, we'll have to modify the scheduling to avoid workflows requesting the API all at the same time.

runningcode:
Thanks. If this doesn't fix the issue, we shouldn't write fixes otherwise Github will automatically close the issue when the PR is merged.
For my understanding before reviewing this, is this meant to be merged to master and released as an update to all users of these github actions? Or do we want to release this to ourselves only for debugging the issue?

jprinet:
My guess is that the Bearer token was not taken into account, therefore defaulting to lower rate limit, we'll see if this is doing the trick.
If rate limiting is the issue, the fix will be to change the scheduling at the consumer side.

jprinet:
Curl retries happen only in case of [transient error](https://everything.curl.dev/usingcurl/downloads/retry), a rate limit exceeded gives [HTTP 403](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limit-http-headers) which does not trigger a retry

jthurne (approved):

.github/actions/gradle/download/action.yml
```diff
@@ -11,22 +11,25 @@ runs:
...
 
-        # Build the command to get the details of the latest version of the build validation scripts
-        cmdGetLatestReleaseData="curl -s $authHeader https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest"
+        # Set curl params to get latest release data of the build validation scripts
+        curlParamsGetLatestReleaseData="$curlOpts https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest"
 
-        # Get the download url of the latest version of the build validation scripts
-        downloadUrl=$($cmdGetLatestReleaseData | jq -c '.assets[] | select(.content_type == "application/zip")' | jq -r .browser_download_url | grep -v maven)
+        # Get the download url of the latest release
+        downloadUrl=$(eval curl $curlParamsGetLatestReleaseData | jq -c '.assets[] | select(.content_type == "application/zip")' | jq -r .browser_download_url | grep -v maven)
```
erichaagdev, line 26: Is `eval` needed? Same question for line 32.

jprinet: This is actually required to keep the quotes around the `Authorization` header. 
I tried to use a quote protection approach which would have been better (`\'` or `\"`) but it resulted unsuccessful as the quote was lost at execution time.
Feel free to suggest an alternative solution 🥹

erichaagdev: Ah, that makes sense! What I would recommend is storing each argument in an array and then expanding the array when `curl` is called. 

We do this _a lot_ throughout the build validation scripts code. For example, see this [`invoke_gradle`](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/b6e9e6d32afba7e41f4491dba2e787f97ea4ebeb/components/scripts/lib/gradle.sh#L46) function where we build out the arguments for calling Gradle.

jprinet: alright, I considered this approach but thought it could create issues on Windows runner, not sure `eval` would behave better though 😅
If this is used inside the scripts, I can definitely use it in the composite action, thanks for the hint, will share something shortly.

erichaagdev: Ah, I have no idea if it would cause any issues with Windows runners. I see we specify `shell: bash` but I don't know what that means when running on Windows.

jthurne: The Windows runners us WSL (Windows Subsystem for Linux) and execute commands using a full and real version of bash. So using arrays should be fine for the Windows runners as well.

jprinet: just confirming as I successfully tested the action on `windows-runner` 😅

erichaagdev (approved):
👍

PR #271, opened 2023-01-03
Fix createReleaseTag ordering

alextu:
This should ensure `createReleaseTag` is executed after `build` whenever `build githubRelease` is requested.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -262,6 +262,8 @@ githubRelease {
 }
 
 val createReleaseTag by tasks.registering(CreateGitTag::class) {
+    // Ensure tag is created only after a successful build
+    mustRunAfter("build")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -262,6 +262,8 @@ githubRelease {
 }
 
 val createReleaseTag by tasks.registering(CreateGitTag::class) {
+    // Ensure tag is created only after a successful build
+    mustRunAfter("build")
```
runningcode, line 266: why did we use a different task here than in the other projects?

alextu: Other projects are Gradle plugins, this one is not and does not publish either to binary repos afaik

runningcode (approved):

PR #272, opened 2023-01-03
Display serialization factor in summary of all experiments

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. #264
5. #265
6. #266
7. #267
8. ⛄ #272
9. #274
10. #283

⛄ = you are here

## Summary

ℹ️ This PR continues the implementation of https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/206. This felt like a good place to stop. More PRs will be coming.

This PR updates the summary of all experiments to display the serialization factor of the first build. For example:

image

## Testing

To test this, I ran the following experiments and verified they displayed the serialization factor.

```sh
./01-validate-incremental-building.sh -r git@github.com:junit-team/junit5.git -t 'assemble' -s https://ge.solutions-team.gradle.com --debug
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:junit-team/junit5.git -t 'assemble' -s https://ge.solutions-team.gradle.com --debug
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x --debug
./03-validate-local-build-caching-different-locations.sh -r git@github.com:junit-team/junit5.git -t 'assemble' -s https://ge.solutions-team.gradle.com --debug
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x --debug
./04-validate-remote-build-caching-ci-ci.sh -1 https://ge.solutions-team.gradle.com/s/atkuswzf5tqbe -2 https://ge.solutions-team.gradle.com/s/gso4oyqjeri4e --debug
./05-validate-remote-build-caching-ci-local.sh -1 https://ge.solutions-team.gradle.com/s/us4sfslrqlfnc -s https://ge.solutions-team.gradle.com --debug
./01-validate-local-build-caching-same-location.sh -r git@github.com:xwiki/xwiki-commons.git -g 'package' -a '-DskipTests' -s https://ge.solutions-team.gradle.com --debug
./01-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/maven/3.8.x/ge -g 'package' -x --debug
[...]

PR #267, opened 2022-12-30
Modernize Java launch code

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. #264
5. #265
6. #266
7. ⛄ #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

This PR updates the Java launch code used to invoke the command line tools. 

Closes #176

## Testing

In order to test this, I ran Gradle experiments 1-3 both online and offline:

```sh
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build'
./01-validate-incremental-building.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build'
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build'
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/ge-solutions.git -p sample-projects/gradle/7.x/ge -t 'build' -x
```

PR #266, opened 2022-12-30
Add variable to query debug mode

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. #264
5. #265
6. ⛄ #266
7. #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

This PR adds a global `debug_mode` variable that can queried throughout the scripts. Previously the variable generated by Argbash was being used. 

Closes #180

PR #265, opened 2022-12-30
Adopt Foojay Toolchains and Kotlin DSL plugins in build logic

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. #264
5. ⛄ #265
6. #266
7. #267
8. #272
9. #274
10. #283

⛄ = you are here

## Summary

This PR updates `buildSrc` to auto-provision the required JVM toolchains using the [Foojay Toolchains Plugin](https://github.com/gradle/foojay-toolchains) and sets the distribution vendor to `AZUL`. 

Additionally, I updated `buildSrc` to use the `kotlin-dsl` plugin. See [this comment for context](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/260#issuecomment-1366705438)

Closes #254

PR #274, opened 2023-01-04
Display potential build time savings in experiment summaries

erichaagdev:
## ❄️ Winter Update PR Navigator ❄️

1. #261
2. #262
3. #263
4. #264
5. #265
6. #266
7. #267
8. #272
9. ⛄ #274
10. #283

⛄ = you are here

## Summary

ℹ️ This PR continues the implementation of https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/206. This felt like a good place to stop. More PRs will be coming.

This PR adds the `potential_build_time` and `potential_build_time_savings`, which is calculated as:

```
potential_build_time = build_2_effective_duration - (build_2_executed_cacheable_duration / build_2_serialization_factor)
potential_build_time_savings = build_1_effective_duration - potential_build_time
```

This is then displayed to the user as in the following example:

image

## Testing

These changes were tested by running the following experiments and verifying they display a potential build time and potential build time savings. Note the exclusion of `exp5-gradle` and `exp4-maven`. I excluded these because I could not easily find a CI build with a cache miss.

```sh
./01-validate-incremental-building.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com
./01-validate-incremental-building.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com -x
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com
./02-validate-local-build-caching-same-location.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com -x
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com
./03-validate-local-build-caching-different-locations.sh -r git@github.com:gradle/androidx.git -p biometric -t 'buildOnServer' -s https://ge.solutions-team.gradle.com -x
./04-validate-remote-build-caching-ci-ci.sh -1 https://ge.solutions-team.gradle.com/s/4jmo6kqqjnebg -2 https://ge.solutions-team.gradle.com/s/5bwb5ofenxrts
./01-validate-local-build-caching-same-location.sh -r git@github.com:OpenAPITools/openapi-generator.git -g 'package' -a '-DskipTests' -e -s https://ge.solutions-team.gradle.com
./01-validate-local-build-caching-same-location.sh -r git@github.com:OpenAPITools/openapi-generator.git -g 'package' -a '-DskipTests' -e -s https://ge.solutions-team.gradle.com -x
./02-validate-local-build-caching-different-locations.sh -r git@github.com:OpenAPITools/openapi-generator.git -g 'package' -a '-DskipTests' -e -s https://ge.solutions-team.gradle.com
./02-validate-local-build-caching-different-locations.sh -r git@github.com:OpenAPITools/openapi-generator.git -g 'package' -a '-DskipTests' -e -s https://ge.solutions-team.gradle.com -x
./03-validate-remote-build-caching-ci-ci.sh -1 https://ge.solutions-team.gradle.com/s/cdfz7q2whfxia -2 https://ge.solutions-team.gradle.com/s/2elplm2ff76am
```

### DIFF
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/Fields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/Fields.java
@@ -33,6 +33,7 @@ public enum Fields {
     EXECUTED_NOT_CACHEABLE_DURATION("Executed not cacheable duration", d -> totalDuration(d, "executed_not_cacheable")),
     EFFECTIVE_TASK_EXECUTION_DURATION("Effective task execution duration", d -> String.valueOf(d.getEffectiveTaskExecutionDuration().toMillis())),
     SERIALIZATION_FACTOR("Serialization factor", d -> formatBigDecimal(d.getSerializationFactor())),
+    EXECUTED_CACHEABLE_DURATION_MILLISECONDS("Executed cacheable duration milliseconds", d -> totalDurationMillis(d, "executed_cacheable")),
     ;
 
     public final String label;
@@ -74,6 +75,15 @@ public enum Fields {
             data.getTasksByAvoidanceOutcome()
                 .getOrDefault(avoidanceOutcome, TaskExecutionSummary.ZERO)
                 .totalDuration()
+        );
+    }
+
+    private static String totalDurationMillis(BuildValidationData data, String avoidanceOutcome) {
+        return String.valueOf(
+            data.getTasksByAvoidanceOutcome()
+                .getOrDefault(avoidanceOutcome, TaskExecutionSummary.ZERO)
+                .totalDuration()
+                .toMillis()
         );
     }
 
--- a/components/scripts/lib/info.sh
+++ b/components/scripts/lib/info.sh
@@ -140,6 +140,8 @@ print_performance_characteristics() {
 
   print_realized_build_time_savings
 
+  print_potential_build_time_savings
+
   print_build_caching_leverage_metrics
 
   print_serialization_factor
@@ -170,6 +172,34 @@ print_realized_build_time_savings() {
     fi
     summary_row "Realized build time savings:" "${value}"
   fi
+}
+
+print_potential_build_time_savings() {
+  local build_1_effective_execution_duration="${effective_task_execution_duration[0]}"
+
+  local build_2_effective_execution_duration="${effective_task_execution_duration[1]}"
+  local build_2_executed_cacheable_duration="${executed_cacheable_duration_milliseconds[1]}"
+  local build_2_serialization_factor="${serialization_factors[1]}"
+
+  # Do not print potential build time savings at all if these values do not exist
+  # This can happen since build-scan-support-tool does not yet support these fields
+  if [[ -z "${build_1_effective_execution_duration}" || -z "${build_2_effective_execution_duration}" || -z "${build_2_executed_cacheable_duration}" || -z "${build_2_serialization_factor}" ]]; then
+    return 0
+  fi
+
+  local value
+  value=""
+  # Only calculate realized build time savings when these have valid values
+  # These values can be returned as zero when an error occurs processing the Build Scan data
+  if [[ "${build_1_effective_execution_duration}" && "${build_2_effective_execution_duration}" && -n "${build_2_serialization_factor}" ]]; then
+    local first_build second_build potential_build_duration potential_savings
+    first_build=$(format_duration "${effective_task_execution_duration[0]}")
+    potential_build_duration=$(echo "${build_2_effective_execution_duration}-(${build_2_executed_cacheable_duration}/${build_2_serialization_factor})" | bc)
+    potential_savings=$(format_duration build_1_effective_execution_duration-potential_build_duration)
+    second_build=$(format_duration potential_build_duration)
+    value="${potential_savings} wall-clock time (from ${first_build} to ${second_build})"
+  fi
+  summary_row "Potential build time savings:" "${value}"
 }
 
 print_build_caching_leverage_metrics() {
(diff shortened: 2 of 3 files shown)


components/scripts/lib/info.sh
```diff
@@ -172,6 +174,35 @@ print_realized_build_time_savings() {
   fi
 }
 
+print_potential_build_time_savings() {
```
erichaagdev, line 177: As this is currently written, the `Potential build time savings` is always displayed. In cases where no cacheable tasks were executed, the metrics are identical to `Realized build time savings`. For example:

image

I wonder if it makes sense to conditionally print `Potential build time savings` only when the number of executed cacheable tasks is > 0?

components/scripts/lib/info.sh
```diff
@@ -140,6 +140,8 @@ print_performance_characteristics() {
 
   print_realized_build_time_savings
 
+  print_potential_build_time_savings
+
```
erichaagdev, line 144: Thoughts on updating the `WARNING` to point out the `Potential build time savings`? 

image

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/Fields.java
```diff
@@ -33,6 +33,7 @@ public enum Fields {
     EXECUTED_NOT_CACHEABLE_DURATION("Executed not cacheable duration", d -> totalDuration(d, "executed_not_cacheable")),
     EFFECTIVE_TASK_EXECUTION_DURATION("Effective task execution duration", d -> String.valueOf(d.getEffectiveTaskExecutionDuration().toMillis())),
     SERIALIZATION_FACTOR("Serialization factor", d -> formatBigDecimal(d.getSerializationFactor())),
+    EXECUTED_CACHEABLE_DURATION_MILLISECONDS("Executed cacheable duration milliseconds", d -> totalDurationMillis(d, "executed_cacheable")),
```
erichaagdev, line 36: I added this as an extra field so that these changes are backwards compatible with the current version of the `build-scan-support-tool`.

I plan to create a follow up PR which removes all special logic from the build validation scripts to use once the `build-scan-support-tool` is updated.

components/scripts/lib/info.sh
jthurne: Ha! I actually suggested tweaking this warning message slightly on one of the earlier PRs to use the "build time savings" language. Is that sort of what you have in mind?

components/scripts/lib/info.sh
jthurne: I think it is better to always display it. I think it would create more confusion to only display it if there were any unrealized build time savings (it would lead users to have to ask themselves why potential build time savings weren't displayed).

jthurne, line 177: 0: Technically, this shows the potential build time savings that could be realized only from build caching and parallel task execution.  It doesn't take into account other potential build time savings, such as those gained using Test Distribution or Predictive Test Selection.

I wonder if we should clarify this somehow. Or make it clearer that, even if the "potential build time savings" have been fully realized that there are still ways in which to further optimize and improve the build.

jthurne (approved):

components/scripts/lib/info.sh
erichaagdev: I will defer this to @etiennestuder.

components/scripts/lib/info.sh
```diff
@@ -172,6 +174,35 @@ print_realized_build_time_savings() {
...
+  local value
+  value=""
+  # Only calculate realized build time savings when these have valid values
+  # These values can be returned as zero when an error occurs processing the Build Scan data
+  if [[ "${build_1_effective_execution_duration}" && "${build_2_effective_execution_duration}" && -n "${build_2_serialization_factor}" ]]; then
+    local first_build second_build potential_build_duration potential_savings
+    first_build=$(format_duration "${effective_task_execution_duration[0]}")
+    # shellcheck disable=SC2034 # it's used on the next few lines
+    potential_build_duration=$(echo "${build_2_effective_execution_duration}-(${build_2_executed_cacheable_duration}/${build_2_serialization_factor})" | bc)
```
jthurne, line 198: Do we know if bc is widely available on many platforms?  I think it probably is, but we should make sure to be safe.

We could use Bash Arithmetic Expansion, but it only supports operations on integers. We could still use it if some floating point percision is necessary, we would just need to multiply the starting values by some multiple of 10, and we could manage the decimal place ourselves.  Only mentioning this if bc is not universal.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/Fields.java
erichaagdev: This has been cleaned up here: https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/283

components/scripts/lib/info.sh
etiennestuder: I would leave the warning text as it is since strictly speaking, we are reducing task exec time and that might not actually translate to build time savings depending on the parallelism of the build.

components/scripts/lib/info.sh
jthurne: Resolving because we came up with a better way to display this which we will address in a future PR.

etiennestuder: We had a longer discussion about this and came up with a proposal that will be implemented in a separate PR.

etiennestuder: We had a longer discussion about this and came up with a proposal that will be implemented in a separate PR.

etiennestuder (approved):

PR #309, opened 2023-02-14
Remove explicit task dependency for Gradle 8

### LINKED ISSUE #308: Avoid explicit dependency on `assembleGradleScripts, assembleMavenScripts`
Introduced here: https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/307/files

clayburn:
Rather than having an explicit dependency from `generateChecksums` to the `assembleScript` Zip tasks, this change declares the two tasks as being inputs to `generateChecksums`

Fixes #308

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -236,14 +236,9 @@ tasks.check {
 val generateChecksums by tasks.registering(Checksum::class) {
     group = "distribution"
     description = "Generates checksums for the distribution zip files."
-    inputFiles.setFrom(assembleGradleScripts.flatMap { gradleScripts: Zip ->
-        assembleMavenScripts.map { mavenScripts: Zip ->
-            gradleScripts.outputs.files.plus(mavenScripts.outputs.files)
-        }
-    })
+    inputFiles.setFrom(assembleGradleScripts, assembleMavenScripts)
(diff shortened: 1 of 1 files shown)


etiennestuder (approved):
LGTM from the code. I did not actually run a build. Please also have @jthurne review it.

build.gradle.kts
```diff
@@ -236,14 +236,9 @@ tasks.check {
 val generateChecksums by tasks.registering(Checksum::class) {
     group = "distribution"
     description = "Generates checksums for the distribution zip files."
-    inputFiles.setFrom(assembleGradleScripts.flatMap { gradleScripts: Zip ->
-        assembleMavenScripts.map { mavenScripts: Zip ->
-            gradleScripts.outputs.files.plus(mavenScripts.outputs.files)
-        }
-    })
+    inputFiles.setFrom(assembleGradleScripts, assembleMavenScripts)
```
jthurne, line 239: 👍 This is much nicer than the old way.

jthurne (approved):
I ran a few test builds and all seems to be well.

build.gradle.kts
clayburn: Thanks @jthurne!

PR #334, opened 2023-03-08
Better handle cases where builds fail because the JVM exits unexpectedly

### LINKED ISSUE #275: Script does not list second build twice when first build fails
When the first build in `01-validate-incremental-building.sh` for Gradle fails due to an OOM and no build scan link was published the build validation scripts will list the second build twice.

See screenshot below with information on the project:

Screenshot 2023-01-04 at 4 26 24 PM

jthurne:
Previously, the build validation scripts depended on the order build scans were
written to the build_scans.csv file, but this led to subtle and weird bugs,
particularly if the JVM running the build exits unexpectedly (such as in
response to an OutOfMemoryError).

This commit reworks things so that each row in the build_scans.csv file is
directly associated with the experiment build that generated it. When reading
build scan data, the association is re-established by matching build scan ids
with those that were produced by the builds.

A few other changes were made so that the build summary output makes sense,
particularly if the first build fails because the JVM exits unexpectedly.

Fixes #275

### DIFF
--- a/components/scripts/lib/maven.sh
+++ b/components/scripts/lib/maven.sh
@@ -54,7 +54,7 @@ invoke_maven() {
 
   args+=(
     -Dmaven.ext.class.path="${extension_classpath}"
-    -Dcom.gradle.enterprise.build_validation.build_number="${build_number}"
+    -Dcom.gradle.enterprise.build_validation.buildNumber="${build_number}"
     -Dcom.gradle.enterprise.build_validation.experimentDir="${EXP_DIR}"
     -Dcom.gradle.enterprise.build_validation.expId="${EXP_SCAN_TAG}"
     -Dcom.gradle.enterprise.build_validation.runId="${RUN_ID}"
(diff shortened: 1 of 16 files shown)


components/scripts/lib/maven.sh
```diff
@@ -52,6 +54,7 @@ invoke_maven() {
 
   args+=(
     -Dmaven.ext.class.path="${extension_classpath}"
+    -Dcom.gradle.enterprise.build_validation.build_number="${build_number}"
```
etiennestuder, line 57: I would rename `build_number` to `buildNumber` (for Gradle and Maven).

etiennestuder (changes requested):
LGTM. One suggestion for change.

components/scripts/lib/maven.sh
jthurne: ✅ Done

etiennestuder (approved):

PR #337, opened 2023-03-09
etst/sys-props

etiennestuder:


components/scripts/lib/gradle.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
 invoke_gradle() {
-  local build_number args
+  local run_num args
   args=()
-  build_number=$1
```
jthurne, line 6: There are a few other places where we reference `build_number` that we may want to rename to `run_num` for consistency. Specifically, the individual experiment scripts have an `execute_build` function that refers to `build_number`.

etiennestuder: Thanks. Fixed.

PR #338, opened 2023-03-09
Empty array elements from additional arguments are no longer added to invocation arguments

erichaagdev:
This is once again an attempt to address the failing cross platform tests. I tested this with the following command and also relied on the cross platform tests:

```
./01-validate-incremental-building.sh \
  -r https://github.com/gradle/common-custom-user-data-gradle-plugin \
  -t build \
  -a '--no-configuration-cache -Pfooo=barr -Pfoo=bar -Porg.gradle.jvmargs=-Xmx1g\ -XX:MaxMetaspaceSize=512m'
```

### DIFF
--- a/components/scripts/lib/maven.sh
+++ b/components/scripts/lib/maven.sh
@@ -67,7 +67,10 @@ invoke_maven() {
 
   # https://stackoverflow.com/a/31485948
   while IFS= read -r -d ''; do
-    args+=( "$REPLY" )
+    local extra_arg="$REPLY"
+    if [ -n "$extra_arg" ]; then
+      args+=("$extra_arg")
+    fi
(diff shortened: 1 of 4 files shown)


components/scripts/lib/maven.sh
```diff
@@ -67,7 +67,10 @@ invoke_maven() {
 
   # https://stackoverflow.com/a/31485948
   while IFS= read -r -d ''; do
-    args+=( "$REPLY" )
+    local extra_arg="$REPLY"
+    if [ -n "$extra_arg" ]; then
+      args+=("$extra_arg")
+    fi
```
jthurne, line 73: You could shorten this to:

```
 [ -n "${extra_arg}" ] && args+=("${extra_arg}")
```

PR #339, opened 2023-03-09
No longer include --scan when displaying the build command

jthurne:
#### Before
image

image

#### After
image

image

### DIFF
--- a/components/scripts/gradle/01-validate-incremental-building.sh
+++ b/components/scripts/gradle/01-validate-incremental-building.sh
@@ -148,11 +148,11 @@ execute_build() {
 }
 
 print_gradle_command() {
-  local scan_arg
-  if [[ "${build_scan_publishing_mode}" == "off" ]]; then
-    scan_arg=" -Dscan.dump"
+  if [[ "${build_scan_publishing_mode}" == "on" ]]; then
+    info "./gradlew --no-build-cache -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
+  else
+    info "./gradlew --no-build-cache -Dscan.dump -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
   fi
-  info "./gradlew --no-build-cache${scan_arg} -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
 }
 
 fetch_build_cache_metrics() {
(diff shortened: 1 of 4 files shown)


components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -149,12 +149,10 @@ execute_build() {
 
 print_gradle_command() {
   local scan_arg
-  if [[ "${build_scan_publishing_mode}" == "on" ]]; then
-    scan_arg="--scan"
-  else
-    scan_arg="-Dscan.dump"
+  if [[ "${build_scan_publishing_mode}" != "on" ]]; then
```
erichaagdev, line 152: It looks like we typically check for `on` vs `off` in the codebase.

```suggestion
  if [[ "${build_scan_publishing_mode}" == "off" ]]; then
```

components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -149,12 +149,10 @@ execute_build() {
...
+    scan_arg=" -Dscan.dump"
   fi
-  info "./gradlew --no-build-cache ${scan_arg} -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
+  info "./gradlew --no-build-cache${scan_arg} -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
```
erichaagdev, line 155:
```suggestion
  info "./gradlew --no-build-cache ${scan_arg} -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} $*$(print_extra_args)"
```

erichaagdev: This was intentional.

components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -149,12 +149,10 @@ execute_build() {
 
 print_gradle_command() {
```
etiennestuder, line 150: I suggest to use the same if/else construct with the almost-duplicated command as in the other experiments.

components/scripts/gradle/01-validate-incremental-building.sh
```diff
@@ -149,12 +149,10 @@ execute_build() {
...
   local scan_arg
```
etiennestuder, line 151: I suggest to use the same if/else construct with the almost-duplicated command as in the other experiments.

etiennestuder (changes requested):

components/scripts/gradle/01-validate-incremental-building.sh
jthurne: Great idea. ✅ Done.

etiennestuder (approved):

erichaagdev (approved):

PR #341, opened 2023-03-09
Workflows no longer run twice

erichaagdev:
This PR updates the workflows so that:

- Build verification only runs on pull requests pointing to `main`.
- Gradle wrapper verification runs on every push (including to PR branches and `main`)

This aligns the workflow conditions with most of the GE Solutions repositories, with the exception of running build verification on pushes to main. Instead, this repository runs build verification as part of the development release.

### DIFF
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -1,6 +1,9 @@
 name: Verify Build
 
-on: [ pull_request, workflow_dispatch ]
+on:
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
 
 jobs:
   verification:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-verification.yml
```diff
@@ -1,10 +1,8 @@
 name: Verify Build
 
 on:
-  push:
-    branches-ignore:
-      - main
   pull_request:
+    branches: [ main ]
```
jthurne, line 5: I know this is the pattern used across the Solutions repos, but it isn't clear to me why we would only want to run build-verification only on PRs targeting `main`.  Is there a downside to running it on all PRs regardless of the branch they target?

etiennestuder (approved):

.github/workflows/build-verification.yml
erichaagdev: I am going to merge this PR as is for now, but this is a great point. I think this warrants a larger team discussion.

PR #346, opened 2023-03-10
User is better informed when GE plugin is not applied explicitly or implicitly

etiennestuder:


components/scripts/lib/gradle.sh
```diff
@@ -59,8 +59,9 @@ invoke_gradle() {
       build_outcomes+=("FAILED")
   fi
 
-  if [ -f "${EXP_DIR}/build-scan-publish-error.txt" ] && [[ "${build_scan_publishing_mode}" == "on" ]]; then
-    die "ERROR: The experiment cannot continue because publishing the build scan failed."
+  if [ -f "${EXP_DIR}/errors.txt" ]; then
+    # Jim: I need your help here to print the content of the file as part of the "die" message
```
etiennestuder, line 63: @jthurne I need your help here.

erichaagdev (approved):

PR #353, opened 2023-03-10
Experiments end early if Build Scan is not published

erichaagdev:
This PR adds a condition to the Gradle and Maven invocation to end the experiment early if a Build Scan was not published for a build.

### DIFF
--- a/components/scripts/lib/gradle.sh
+++ b/components/scripts/lib/gradle.sh
@@ -64,6 +64,11 @@ invoke_gradle() {
     die "ERROR: The experiment cannot continue because of a non-recoverable failure while running the build: $(cat "${EXP_DIR}/errors.txt")"
   fi
 
+  if is_build_scan_metadata_missing "$run_num"; then
+    print_bl
+    die "ERROR: The experiment cannot continue because a Build Scan was not published."
+  fi
+
   # defined in git.sh
   read_git_metadata_from_current_repo
   requested_tasks+=("${tasks}")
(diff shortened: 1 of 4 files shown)


jthurne (approved):

components/scripts/lib/gradle.sh
```diff
@@ -64,6 +64,11 @@ invoke_gradle() {
     die "ERROR: The experiment cannot continue because of a non-recoverable failure while running the build: $(cat "${EXP_DIR}/errors.txt")"
   fi
 
+  if is_build_scan_metadata_missing "$run_num"; then
+    print_bl
+    die "ERROR: The experiment cannot continue because a Build Scan was not published."
+  fi
```
etiennestuder, line 70: Is this only relevant in online mode? (I'm asking since we are referencing a method in the build-scan-online.sh script)

What happens in offline mode?

components/scripts/lib/gradle.sh
```diff
@@ -64,6 +64,11 @@ invoke_gradle() {
     die "ERROR: The experiment cannot continue because of a non-recoverable failure while running the build: $(cat "${EXP_DIR}/errors.txt")"
   fi
```
etiennestuder, line 66: Did we forget to also add this for Maven? (not strictly related to this PR)

etiennestuder (changes requested):

components/scripts/lib/gradle.sh
erichaagdev: Nice catch. Addressed in 537856e9c6d20a2bbdf276c53ed4bd545667e9a9.

components/scripts/lib/gradle.sh
erichaagdev: Maven does not write to or create an `errors.txt` file so there is nothing to check for.

etiennestuder: Got it.

etiennestuder (approved):

PR #365, opened 2023-03-17
Experiments can be run with parallel GC enabled and configuration cache is not invalidated between runs

erichaagdev:
This PR is a second attempt at PR #361. It's description is still entirely valid. 

However, it introduced new issue for projects using configuration caching. It was since reverted from `main`.

This PR solves the original issue in a way that is compatible with configuration caching.

### DIFF
--- a/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
@@ -12,7 +12,7 @@ initscript {
 
     def getInputParam = { String name ->
         def envVarName = name.toUpperCase().replace('.', '_').replace('-', '_')
-        return System.getProperty(name) ?: System.getenv(envVarName)
+        return gradle.startParameter.systemPropertiesArgs[name] ?: System.getenv(envVarName)
     }
 
     def pluginRepositoryUrl = getInputParam('com.gradle.enterprise.build-validation.gradle.plugin-repository.url')
@@ -61,7 +61,7 @@ if (!isTopLevelBuild) {
 
 def getInputParam = { String name ->
     def envVarName = name.toUpperCase().replace('.', '_').replace('-', '_')
-    return System.getProperty(name) ?: System.getenv(envVarName)
+    return gradle.startParameter.systemPropertiesArgs[name] ?: System.getenv(envVarName)
 }
 
 def geUrl = getInputParam('com.gradle.enterprise.build-validation.gradle-enterprise.url')
--- a/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
@@ -72,7 +72,6 @@ def ccudPluginVersion = getInputParam('com.gradle.enterprise.build-validation.cc
 def expDir = getInputParam('com.gradle.enterprise.build-validation.expDir')
 def expId = getInputParam('com.gradle.enterprise.build-validation.expId')
 def runId = getInputParam('com.gradle.enterprise.build-validation.runId')
-def runNum = getInputParam('com.gradle.enterprise.build-validation.runNum')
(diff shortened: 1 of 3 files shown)


components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -12,7 +12,7 @@ initscript {
 
     def getInputParam = { String name ->
         def envVarName = name.toUpperCase().replace('.', '_').replace('-', '_')
-        return System.getProperty(name) ?: System.getenv(envVarName)
+        return gradle.startParameter.systemPropertiesArgs[name] ?: System.getenv(envVarName)
```
etiennestuder, line 15: Add a comment to this method explaining that the way we read system properties here means they are not tracked as inputs to the configuration cache model. The same is the case for reading environment variables with Gradle [6.5, 7.3].

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -61,7 +61,7 @@ if (!isTopLevelBuild) {
 
 def getInputParam = { String name ->
     def envVarName = name.toUpperCase().replace('.', '_').replace('-', '_')
-    return System.getProperty(name) ?: System.getenv(envVarName)
+    return gradle.startParameter.systemPropertiesArgs[name] ?: System.getenv(envVarName)
```
etiennestuder, line 64: Add the same comment to the method as above.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
etiennestuder, line 15: Add a comment to this method explaining that the way we read system properties here means they are not tracked as inputs to the configuration cache model. The same is the case for reading environment variables with Gradle [6.5, 7.3].

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -72,7 +72,6 @@ def ccudPluginVersion = getInputParam('com.gradle.enterprise.build-validation.cc
 def expDir = getInputParam('com.gradle.enterprise.build-validation.expDir')
 def expId = getInputParam('com.gradle.enterprise.build-validation.expId')
 def runId = getInputParam('com.gradle.enterprise.build-validation.runId')
-def runNum = getInputParam('com.gradle.enterprise.build-validation.runNum')
```
etiennestuder, line 75: Add a comment above this block (and the one above) explaining that these variables do not change between the two invocations of the build; thus, they do not need to be tracked as inputs to the configuration model.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -87,6 +86,11 @@ if (ccudPluginVersion && isNotAtLeast(ccudPluginVersion, '1.7')) {
 def registerBuildScanActions = { def buildScan ->
     def scanFile = new File(expDir, 'build-scans.csv')
     buildScan.buildScanPublished { publishedBuildScan ->
+        def getInputParamCompatibleWithCC = { String name ->
```
etiennestuder, line 89: The name of this method is a bit misleading/unclear. I suggest renaming it to `getRunNumInputParam`.

etiennestuder, line 89: Add a comment to this closure explaining that we defer reading the `runNum` system property to the execution time since that property does not affect the configuration of the build but changes between multiple build invocations. Reading a changing system property value at configuration time in a way that it is not tracked as an input of the configuration model (which we cannot guarantee due to the -XX cases) would mean that for configuration cache-compatible builds, the value determined in the first build would be reused in the second build.

etiennestuder (changes requested):

PR #372, opened 2023-03-21
Stop enforcing that -s must be specified when -e is specified

etiennestuder:


jthurne (approved):

erichaagdev (approved):
I tested these changes for all permutations of `-e` / `-s`, for Gradle versions 5.x and 6.x, for projects with and without Gradle Enterprise already configured, and interactive / non-interactive.

All outcomes are as expected, but there is one strange behavior that is worth reporting. 

When running on a project where Gradle Enterprise is not already configured, and when `-e` is used only (so no `-s`), and on Gradle 5.x, then we have the following outcome:

```
* What went wrong:
A problem occurred configuring root project 'gradle-sample'.
> The Gradle Enterprise server URL is not configured.
  Either configure it directly (see https://docs.gradle.com/enterprise/gradle-plugin/#gradle_5_x_2) in the project,
  or use `--gradle-enterprise-server` when running the build validation script.
> Cannot invoke method endsWith() on null object
```

I'm not sure why we see both error messages. My guess is that throwing an exception inside of the `afterResolve` does not actually fail the build, but I did not verify that assumption. Previously the exception was being thrown outside of the `afterResolve` inside of the `afterEvaluate`.

I will approve the PR because everything else works as expected.

etiennestuder:
Thanks. Good catch, @erichaagdev. I was able to reproduce and fix it. Please take another look. Nothing changed for Gradle >=6.x

PR #373, opened 2023-03-21
Add support for using a license when running in offline mode

jthurne:
- Pass the license file to build-scan-support-tool
- Update Build Scan dump extraction info messages to match those when fetching Build Scan data from Gradle Enterprise
- Immediately die if extracting Build Scan dumps fail


jthurne:
**Successful run**
image

**License File Missing**
image

**License file cannot be read**
image

**License file is invalid**
image

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,20 @@ find_build_scan_dump() {
...
+  args+=(
+      "extract"
+      "--license-file" "${SCRIPT_DIR}/gradle-enterprise.license"
+      "1,${build_scan_dumps[1]}"
+      "0,${build_scan_dumps[0]}"
+  )
+
+  echo "Extracting Build Scan data for all builds"
+  if ! build_scan_csv="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" "${args[@]}")"; then
+    exit "$UNEXPECTED_ERROR"
```
etiennestuder, line 49: exit or die?

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,20 @@ find_build_scan_dump() {
...
 
 read_build_scan_dumps() {
-  local build_scan_csv
-  echo -n "Extracting build scan data"
-  build_scan_csv="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" extract "0,${build_scan_dumps[0]}"  "1,${build_scan_dumps[1]}")"
+  local build_scan_csv args
+  args=()
+
+  args+=(
+      "extract"
```
etiennestuder, line 41: Why do we have to pass `extract`?

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,20 @@ find_build_scan_dump() {
...
+      "--license-file" "${SCRIPT_DIR}/gradle-enterprise.license"
+      "1,${build_scan_dumps[1]}"
+      "0,${build_scan_dumps[0]}"
+  )
```
etiennestuder, line 45: Why 1 and then 0 ?

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,20 @@ find_build_scan_dump() {
...
-  echo -n "Extracting build scan data"
-  build_scan_csv="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" extract "0,${build_scan_dumps[0]}"  "1,${build_scan_dumps[1]}")"
+  local build_scan_csv args
+  args=()
+
+  args+=(
+      "extract"
+      "--license-file" "${SCRIPT_DIR}/gradle-enterprise.license"
+      "1,${build_scan_dumps[1]}"
+      "0,${build_scan_dumps[0]}"
```
etiennestuder, line 44: Do we use the same approach for passing args with the jar file that fetches build scans remotely?

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,20 @@ find_build_scan_dump() {
...
+  )
+
+  echo "Extracting Build Scan data for all builds"
+  if ! build_scan_csv="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" "${args[@]}")"; then
+    exit "$UNEXPECTED_ERROR"
+  fi
   parse_build_scan_csv "$build_scan_csv" "build_cache_metrics_only"
```
etiennestuder, line 51: I suggest we extend the task in our todo doc to also split this up between reading the data (from local disk) and then parsing the data.

etiennestuder (changes requested):

components/scripts/lib/build-scan-offline.sh
erichaagdev: Added.

components/scripts/lib/build-scan-offline.sh
jthurne: Unfortunately we can't use `die` here because we don't have an error message (the offline tool takes care of printing the error messages).  We could call it, but then we would end up with an extra empty line after the scripts terminate.

In fact, I originally used `die` here, but switched it to `exit` when Eric pointed out the extra newline.

components/scripts/lib/build-scan-offline.sh
jthurne: This is a consequence of the CLI framework the offline tool uses. It requires there to be a subcommand, and so Sorin decided to name the subcommand "extract".

components/scripts/lib/build-scan-offline.sh
jthurne: Yes, although for the online fetch, this is generated more dynamically. See https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/422822eec0c2919423a44a5d36eff1c175b0daee/components/scripts/lib/build-scan-online.sh#L100-L102

components/scripts/lib/build-scan-offline.sh
jthurne: ✅ Heh, this was a mistake. Fixed.

erichaagdev (approved):

components/scripts/lib/build-scan-offline.sh
etiennestuder: Could we do the same here as we do in the online case to make the code more similar?

components/scripts/lib/build-scan-offline.sh
etiennestuder: Thanks for the clarification.

components/scripts/lib/build-scan-offline.sh
etiennestuder, line 1: I would add a check that the license file exists in case offline mode is requested. I know we do it in the jar file, too, and I would keep it there, but I would also do it here. Maybe at the time that the cmd line args are evaluated?

etiennestuder (changes requested):

components/scripts/lib/build-scan-offline.sh
etiennestuder: Also, this way, we do not run the builds just to tell the user then that the license file is missing.

components/scripts/lib/build-scan-offline.sh
etiennestuder: If this always involves fetching two build scans, I suggest we keep the current code here and adjust the code for the online mode also to pass on two explicit build scans rather than _n_ build scans.

etiennestuder: I now see that in online mode, we invoke either one or two builds and thus need the generic looping.

I'm withdrawing my suggestion.

etiennestuder (changes requested):

components/scripts/lib/build-scan-offline.sh
etiennestuder: Message:

`ERROR: Missing required file gradle-enterprise.license in the root folder of the build validation scripts`

etiennestuder (changes requested):

components/scripts/lib/build-scan-offline.sh
etiennestuder: I've already adjust the error message for the missing jar file ont his branch:
https://github.com/gradle/gradle-enterprise-build-validation-scripts/commit/70ac1ab9946ce12d0b7609a7fcbadbcbdb02ec4e

etiennestuder: @erichaagdev will you take care of the additional check for the license file or @jthurne ?

erichaagdev: image

erichaagdev: Done in 7f34b23e73a496308adf27be112dade0938b9bf0

jthurne (commented):
Eric's final updates LGTM!

PR #374, opened 2023-03-21
etst/maven-no-server

etiennestuder:


erichaagdev:
> I tried to write to the errors.txt file when the server == null but this either did not work or the shell scripts did not pick it up (do they even look for it?)

I verified the `invoke_maven` function does not check for the `errors.txt` file. I think it makes sense to add this since we now have a use case for it.

> What we could do is that in the configure callback we set the server URL to some non-null dummy value if the value is currently null - this way, we can at least avoid publishing to scans.gradle.com -- if we want this.

This makes sense. It's unfortunate we have to implement it in this way but I think it should be done. It is worth adding if it means avoiding accidentally publishing to scans.gradle.com.

erichaagdev:
> set the server URL to some non-null dummy value

We could use `https://0.0.0.0` which is consistent with what I've done [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/375/files) and with what we used to do [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/245bad0f239b877af55dff5b18eaf35ff64a9229/components/scripts/lib/gradle.sh#L22-L24C5).

etiennestuder:
> I verified the `invoke_maven` function does not check for the `errors.txt` file. I think it makes sense to add this since we now have a use case for it.

@erichaagdev can you please add that check as part of this PR? I'll look into the Java side.

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/CaptureBuildScansListener.java
```diff
@@ -74,14 +80,26 @@ private static String urlEncode(String str) {
 
     private void capturePublishedBuildScan(BuildScanApi buildScan) {
         buildScan.buildScanPublished(scan -> {
+            // abort with an error message written to the error file if no GE server URL is available
+            if (buildScan.getServer() == null || buildScan.getServer() .equals("https://unknown.server.com")) {
```
erichaagdev, line 84: From what I can tell in my local testing, it's much to late to check this here as a Build Scan will not be published if the GE URL is not set.

erichaagdev:
@etiennestuder I have added the `errors.txt` check for Maven. 

image

Note that in order to test and get that output, I had to move the null server check much earlier, which is what prompted [this comment](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/374/files#r1145361984).

components/capture-build-scan-url-maven-extension/src/main/resources/META-INF/plexus/components.xml
```diff
@@ -2,9 +2,9 @@
   <components>
     <component>
       <role>com.gradle.maven.extension.api.GradleEnterpriseListener</role>
-      <role-hint>capture-build-scan-url</role-hint>
-      <implementation>com.gradle.CaptureBuildScansListener</implementation>
-      <description>Captures the build scan URL from Maven builds</description>
+      <role-hint>configure-gradle-enterprise</role-hint>
```
erichaagdev, line 5: Should we also rename the Maven extension project from `capture-build-scan-url-maven-extension` to `configure-gradle-enterprise-maven-extension`? We would also need to update the name of the jar on the classpath when invoking Maven. Without it, this renaming feels only half complete, but I am also fine leaving it.

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/ConfigureGradleEnterprise.java
```diff
@@ -0,0 +1,129 @@
...
+import java.nio.file.Files;
+
+import static java.lang.Boolean.parseBoolean;
+import static java.nio.file.StandardOpenOption.*;
+
+@SuppressWarnings("unused")
+public class ConfigureGradleEnterprise implements GradleEnterpriseListener {
+
+    private static final String EXPERIMENT_DIR = System.getProperty("com.gradle.enterprise.build-validation.expDir");
+    private static final String UNKNOWN_SERVER = "https://unknown.server.org";
```
jthurne, line 25: `unknown.server.org` is a viable domain name. It doesn't resolve to anything today, but in theory it could resolve to something in the future. It also means that a DNS lookup against this address will take some time (to determine that it doesn't exist). 

I'm not sure that matters too much in this case, given how we're using this currently.  But we should be careful how we use this value in the future.

components/capture-build-scan-url-maven-extension/src/main/resources/META-INF/plexus/components.xml
etiennestuder: +1. but let's do after the merge to avoid too many changes. (for the list of 2.3.2)

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/ConfigureGradleEnterprise.java
```diff
@@ -0,0 +1,129 @@
...
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static void append(File file, String text) {
+        try {
+            Files.write(file.toPath(), text.getBytes(), CREATE, WRITE, APPEND);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to write to file to build-scans.csv: " + e.getMessage(), e);
```
jthurne, line 125: This error message specifically references `build-scans.csv`, but we also use this function to append to `errors.txt`. Maybe something like this would make the message generic:

```suggestion
            throw new RuntimeException(String.format("Unable to write to %s: %s", file.getName(), e.getMessage(), e);
```

etiennestuder: An oversight. I'll fix it.

components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/ConfigureGradleEnterprise.java
jthurne: One option is `https://unknown.example.com` or anything ending with `example.com`. The `example.com` domain is reserved for this type of usage.

I think we were originally using `https://0.0.0.0`, but that's not very descriptive. It does avoid a potential DNS lookup though.

jthurne: > I had tried https://0.0.0.0 previously and it will actually still attempt to publish. 

That's surprising since we use this when we pass the flag to generate a build scan dump instead of publishing one. I imagine we would have that problem with any URL we pass here though.

erichaagdev: When we generate a Build Scan dump then publishing is disabled, thus nothing will not attempt to communicate with `https://0.0.0.0`.

For reference, this is what it looks like when using `https://0.0.0.0`:

image

I may actually have a good solution though. In the `buildFinished` callback where we are validating the server, if the server is unset, then we can disable Build Scan publishing. This avoids any DNS calls, timeouts, and the dummy URL being displayed in any user facing text.

I just tried it and here is the outcome:

image

Where currently it looks like:

image

etiennestuder: I suggest we use https://unknown.example.com/. The only scenario where we run into this is when a project has the GE  extension configured but not server defined in the extension.xml.

erichaagdev: Can we just disable Build Scan publishing if the Gradle Enterprise server is null? Then no dummy URL is needed.

erichaagdev: Discussed with @etiennestuder and we decided to disable Build Scan publishing if no Gradle Enterprise is configured. This both avoids publishing to scans.gradle.com and also removes any dummy URLs.

erichaagdev:
Decided to run cross platform tests before merging. 

The run was successful. ✅ 

https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4504350331

PR #379, opened 2023-03-24
Maven extension is renamed to better reflect the latest enhancements

erichaagdev:
This PR renames the Maven extension from `capture-build-scan-url-maven-extension` to `configure-gradle-enterprise-maven-extension`.

In `2.3.2`, many enhancements had been made to this extension that make it do more than just capture a Build Scan. Instead, it now configure's Gradle Enterprise in much the same way the Gradle init script does.


erichaagdev:
Successful cross-platform test run: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4514488159

jthurne (approved):

etiennestuder (approved):

PR #382, opened 2023-03-27
Performance characteristics includes the serialization factors of both builds

erichaagdev:
This PR updates the serialization factor row in the performance characteristics summary to include the serialization factor from both builds. For example:

image


etiennestuder (approved):
LGTM.

PR #383, opened 2023-03-28
Summary calculations are migrated to Java

erichaagdev:
This PR migrates the summary calculation code for build time metrics from Bash to Java. No functional changes have been made.

Additionally, this PR addresses an issue with rendering negative build time savings.

**Before**

image

**After**

image

Successful cross-platform tests run: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4559844622

In order to test these changes, I took the results of the cross platform tests and compared them to the results from a previous version of the scripts (using the CI/CI experiments).

### DIFF
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
@@ -10,7 +10,7 @@ import com.gradle.enterprise.api.model.MavenAttributes;
 import com.gradle.enterprise.api.model.MavenBuildCachePerformance;
 import com.gradle.enterprise.api.model.MavenBuildCachePerformanceGoalExecutionEntry;
 import com.gradle.enterprise.cli.ConsoleLogger;
-import com.gradle.enterprise.model.BuildValidationData;
+import com.gradle.enterprise.model.BuildScanData;
 import com.gradle.enterprise.model.CustomValueNames;
 import com.gradle.enterprise.model.NumberedBuildScan;
 import com.gradle.enterprise.model.TaskExecutionSummary;
@@ -122,7 +122,7 @@ public class GradleEnterpriseApiClient {
         }
     }
 
-    public BuildValidationData fetchBuildValidationData(NumberedBuildScan buildScan) {
+    public BuildScanData fetchBuildValidationData(NumberedBuildScan buildScan) {
         int runNum = buildScan.runNum();
         String buildScanId = buildScan.buildScanId();
         try {
@@ -131,7 +131,7 @@ public class GradleEnterpriseApiClient {
                 GradleAttributes attributes = apiClient.getGradleAttributes(buildScanId, null);
                 GradleBuildCachePerformance buildCachePerformance = apiClient.getGradleBuildCachePerformance(buildScanId, null);
 
-                return new BuildValidationData(
+                return new BuildScanData(
                     runNum,
                     attributes.getRootProjectName(),
                     buildScanId,
@@ -151,7 +151,7 @@ public class GradleEnterpriseApiClient {
                 MavenAttributes attributes = apiClient.getMavenAttributes(buildScanId, null);
                 MavenBuildCachePerformance buildCachePerformance = apiClient.getMavenBuildCachePerformance(buildScanId, null);
 
-                return new BuildValidationData(
+                return new BuildScanData(
                     runNum,
                     attributes.getTopLevelProjectName(),
                     buildScanId,
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
@@ -122,7 +122,7 @@ public class GradleEnterpriseApiClient {
         }
     }
 
-    public BuildScanData fetchBuildValidationData(NumberedBuildScan buildScan) {
+    public BuildScanData fetchBuildScanData(NumberedBuildScan buildScan) {
         int runNum = buildScan.runNum();
         String buildScanId = buildScan.buildScanId();
         try {
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
@@ -2,7 +2,9 @@ package com.gradle.enterprise.cli;
 
 import com.gradle.enterprise.model.BuildValidationData;
 
+import java.util.Arrays;
 import java.util.function.Function;
+import java.util.stream.Stream;
 
 public enum BuildTimeMetricsFields {
     // The order the enums are defined controls the order the fields are printed in the CSV
@@ -19,5 +21,9 @@ public enum BuildTimeMetricsFields {
     BuildTimeMetricsFields(String label, Function<BuildValidationData, String> value) {
         this.label = label;
         this.value = value;
+    }
+
+    public static Stream<BuildTimeMetricsFields> ordered() {
+        return Arrays.stream(BuildTimeMetricsFields.values());
     }
 }
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
@@ -1,29 +1,48 @@
 package com.gradle.enterprise.cli;
 
-import com.gradle.enterprise.model.BuildValidationData;
+import com.gradle.enterprise.model.BuildTimeMetricsData;
 
+import java.time.Duration;
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
 public enum BuildTimeMetricsFields {
     // The order the enums are defined controls the order the fields are printed in the CSV
-    INITIAL_BUILD_TIME("Initial Build Time", d -> ""),
-    INSTANT_SAVINGS("Instant Savings", d -> ""),
-    INSTANT_SAVINGS_BUILD_TIME("Instant Savings Build Time", d -> ""),
-    PENDING_SAVINGS("Pending Savings", d -> ""),
-    PENDING_SAVINGS_BUILD_TIME("Pending Savings Build Time", d -> ""),
+    INITIAL_BUILD_TIME("Initial Build Time", d -> formatDuration(d.getInitialBuildTime())),
+    INSTANT_SAVINGS("Instant Savings", d -> formatDuration(d.getInstantSavings())),
+    INSTANT_SAVINGS_BUILD_TIME("Instant Savings Build Time", d -> formatDuration(d.getInstantSavingsBuildTime())),
+    PENDING_SAVINGS("Pending Savings", d -> formatDuration(d.getPendingSavings())),
+    PENDING_SAVINGS_BUILD_TIME("Pending Savings Build Time", d -> formatDuration(d.getPendingSavingsBuildTime())),
     ;
 
     public final String label;
-    public final Function<BuildValidationData, String> value;
+    public final Function<BuildTimeMetricsData, String> value;
 
-    BuildTimeMetricsFields(String label, Function<BuildValidationData, String> value) {
+    BuildTimeMetricsFields(String label, Function<BuildTimeMetricsData, String> value) {
         this.label = label;
         this.value = value;
     }
 
     public static Stream<BuildTimeMetricsFields> ordered() {
         return Arrays.stream(BuildTimeMetricsFields.values());
+    }
+
+    private static String formatDuration(Duration duration) {
+        long hours = duration.toHours();
+        long minutes = duration.minusHours(hours).toMinutes();
+        double seconds = duration.minusHours(hours).minusMinutes(minutes).toMillis() / 1000d;
+
+        StringBuilder s = new StringBuilder();
+        if (hours != 0) {
+            s.append(hours).append("h ");
+        }
+        if (minutes != 0) {
+            s.append(minutes).append("m ");
+        }
+        s.append(String.format(Locale.ROOT, "%.3fs", seconds));
+
+        return s.toString().trim();
     }
 }
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
@@ -10,11 +10,11 @@ import java.util.stream.Stream;
 
 public enum BuildTimeMetricsFields {
     // The order the enums are defined controls the order the fields are printed in the CSV
-    INITIAL_BUILD_TIME("Initial Build Time", d -> formatDuration(d.getInitialBuildTime())),
-    INSTANT_SAVINGS("Instant Savings", d -> formatDuration(d.getInstantSavings())),
-    INSTANT_SAVINGS_BUILD_TIME("Instant Savings Build Time", d -> formatDuration(d.getInstantSavingsBuildTime())),
-    PENDING_SAVINGS("Pending Savings", d -> formatDuration(d.getPendingSavings())),
-    PENDING_SAVINGS_BUILD_TIME("Pending Savings Build Time", d -> formatDuration(d.getPendingSavingsBuildTime())),
+    INITIAL_BUILD_TIME("Initial Build Time", d -> formatDuration(d.initialBuildTime)),
+    INSTANT_SAVINGS("Instant Savings", d -> formatDuration(d.instantSavings)),
+    INSTANT_SAVINGS_BUILD_TIME("Instant Savings Build Time", d -> formatDuration(d.instantSavingsBuildTime)),
+    PENDING_SAVINGS("Pending Savings", d -> formatDuration(d.pendingSavings)),
+    PENDING_SAVINGS_BUILD_TIME("Pending Savings Build Time", d -> formatDuration(d.pendingSavingsBuildTime)),
     ;
 
     public final String label;
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
@@ -1,0 +1,23 @@
+package com.gradle.enterprise.cli;
+
+import com.gradle.enterprise.model.BuildValidationData;
+
+import java.util.function.Function;
+
+public enum BuildTimeMetricsFields {
+    // The order the enums are defined controls the order the fields are printed in the CSV
+    INITIAL_BUILD_TIME("Initial Build Time", d -> ""),
+    INSTANT_SAVINGS("Instant Savings", d -> ""),
+    INSTANT_SAVINGS_BUILD_TIME("Instant Savings Build Time", d -> ""),
+    PENDING_SAVINGS("Pending Savings", d -> ""),
+    PENDING_SAVINGS_BUILD_TIME("Pending Savings Build Time", d -> ""),
+    ;
+
+    public final String label;
+    public final Function<BuildValidationData, String> value;
+
+    BuildTimeMetricsFields(String label, Function<BuildValidationData, String> value) {
+        this.label = label;
+        this.value = value;
+    }
+}
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
@@ -1,6 +1,6 @@
 package com.gradle.enterprise.cli;
 
-import com.gradle.enterprise.model.BuildTimeMetricsData;
+import com.gradle.enterprise.model.BuildTimeMetrics;
 
 import java.time.Duration;
 import java.util.Arrays;
@@ -18,9 +18,9 @@ public enum BuildTimeMetricsFields {
     ;
 
     public final String label;
-    public final Function<BuildTimeMetricsData, String> value;
+    public final Function<BuildTimeMetrics, String> value;
 
-    BuildTimeMetricsFields(String label, Function<BuildTimeMetricsData, String> value) {
+    BuildTimeMetricsFields(String label, Function<BuildTimeMetrics, String> value) {
         this.label = label;
         this.value = value;
     }
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
@@ -9,7 +9,7 @@ import java.util.Arrays;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
-public enum BuildValidationFields {
+public enum BuildValidationDataFields {
     // The order the enums are defined controls the order the fields are printed in the CSV
     RUN_NUM("Run Num", d -> toStringSafely(d.runNum())),
     ROOT_PROJECT_NAME("Root Project Name", BuildValidationData::getRootProjectName),
@@ -40,13 +40,13 @@ public enum BuildValidationFields {
     public final String label;
     public final Function<BuildValidationData, String> value;
 
-    BuildValidationFields(String label, Function<BuildValidationData, String> value) {
+    BuildValidationDataFields(String label, Function<BuildValidationData, String> value) {
         this.label = label;
         this.value = value;
     }
 
-    public static Stream<BuildValidationFields> ordered() {
-        return Arrays.stream(BuildValidationFields.values());
+    public static Stream<BuildValidationDataFields> ordered() {
+        return Arrays.stream(BuildValidationDataFields.values());
     }
 
     private static String toStringSafely(Object object) {
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
@@ -3,7 +3,7 @@ package com.gradle.enterprise.cli;
 import com.gradle.enterprise.api.client.FailedRequestException;
 import com.gradle.enterprise.api.client.GradleEnterpriseApiClient;
 import com.gradle.enterprise.model.BuildTimeMetrics;
-import com.gradle.enterprise.model.BuildValidationData;
+import com.gradle.enterprise.model.BuildScanData;
 import com.gradle.enterprise.model.CustomValueNames;
 import com.gradle.enterprise.model.NumberedBuildScan;
 import com.gradle.enterprise.network.NetworkSettingsConfigurator;
@@ -25,12 +25,12 @@ import java.util.stream.Collectors;
     mixinStandardHelpOptions = true,
     description = "Fetches data relevant to build validation from the given build scans."
 )
-public class FetchBuildValidationDataCommand implements Callable<Integer> {
+public class FetchBuildScanDataCommand implements Callable<Integer> {
 
     private final CommandLine.Help.ColorScheme colorScheme;
     private ConsoleLogger logger;
 
-    public FetchBuildValidationDataCommand(CommandLine.Help.ColorScheme colorScheme) {
+    public FetchBuildScanDataCommand(CommandLine.Help.ColorScheme colorScheme) {
         this.colorScheme = colorScheme;
     }
 
@@ -66,35 +66,35 @@ public class FetchBuildValidationDataCommand implements Callable<Integer> {
             .orElse(CustomValueNames.DEFAULT);
 
         logStartFetchingAllBuildScanData();
-        List<BuildValidationData> buildValidationData = fetchBuildScanData(buildScans, customValueKeys);
+        List<BuildScanData> buildScanData = fetchBuildScanData(buildScans, customValueKeys);
 
         logFinishedFetchingAllBuildScanData();
-        logFetchResults(buildValidationData, customValueKeys);
+        logFetchResults(buildScanData, customValueKeys);
 
-        printResults(buildValidationData);
+        printResults(buildScanData);
 
         return ExitCode.OK;
     }
 
     @NotNull
-    private List<BuildValidationData> fetchBuildScanData(List<NumberedBuildScan> buildScans, CustomValueNames customValueKeys) {
+    private List<BuildScanData> fetchBuildScanData(List<NumberedBuildScan> buildScans, CustomValueNames customValueKeys) {
         return buildScans.stream()
             .parallel()
             .map(buildScan -> fetchBuildScanData(buildScan, customValueKeys))
             .collect(Collectors.toList());
     }
 
-    private BuildValidationData fetchBuildScanData(NumberedBuildScan buildScan, CustomValueNames customValueNames) {
+    private BuildScanData fetchBuildScanData(NumberedBuildScan buildScan, CustomValueNames customValueNames) {
         logStartFetchingBuildScanData(buildScan);
         try {
             GradleEnterpriseApiClient apiClient = new GradleEnterpriseApiClient(buildScan.baseUrl(), customValueNames, logger);
-            BuildValidationData data = apiClient.fetchBuildValidationData(buildScan);
+            BuildScanData data = apiClient.fetchBuildValidationData(buildScan);
 
             logFinishedFetchingBuildScanData(buildScan);
             return data;
         } catch (RuntimeException e) {
             logException(e);
-            return new BuildValidationData(
+            return new BuildScanData(
                 buildScan.runNum(),
                 "",
                 buildScan.buildScanId(),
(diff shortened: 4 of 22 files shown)


components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public static BuildTimeMetrics from(BuildValidationData firstBuild, BuildValidationData secondBuild) {
+        final Duration buildTimeFirstBuild = firstBuild.getBuildTime();
+        final Duration buildTimeSecondBuild = secondBuild.getBuildTime();
+        final TaskExecutionSummary executedCacheableTaskSummarySecondBuild = secondBuild.getExecutedCacheableSummary();
+        final BigDecimal serializationFactorSecondBuild = secondBuild.getSerializationFactor();
+
+        if (buildTimeFirstBuild == null || buildTimeSecondBuild == null || executedCacheableTaskSummarySecondBuild == null || serializationFactorSecondBuild == null) {
+            return null;
+        }
```
erichaagdev, line 35: I'm not sure about this `null` checking. It's unclear if the Gradle Enterprise API could return these as `null`. We also we deliberately return [these fields as `null` from the API client](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/3870d12078a7410dceb439b53085ebed6cb63c98/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java#L101-L114) if anything goes wrong.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public final String label;
+    public final Function<BuildTimeMetrics, String> value;
+
+    BuildTimeMetricsFields(String label, Function<BuildTimeMetrics, String> value) {
+        this.label = label;
+        this.value = value;
+    }
+
+    public static Stream<BuildTimeMetricsFields> ordered() {
```
etiennestuder, line 28: Returning a Stream is a bit odd, I find. I would return a `List` instead.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
```diff
@@ -0,0 +1,48 @@
...
+        return Arrays.stream(BuildTimeMetricsFields.values());
+    }
+
+    private static String formatDuration(Duration duration) {
```
etiennestuder, line 32: Isn't there a formatter built-in to Java to render durations?

A quick search reveals nothing. But I don't think you need to do any math:

System.out.println(String.format("%sh %sm %ss", 
                d.toHoursPart(), 
                d.toMinutesPart(), 
                d.toSecondsPart()));

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
```diff
@@ -70,9 +71,12 @@ public Integer call() {
         logFinishedFetchingAllBuildScanData();
         logFetchResults(buildValidationData, customValueKeys);
 
-        printHeader();
+        printBuildValidationHeader();
         printBuildValidationData(buildValidationData);
 
+        printBuildTimeMetricsHeader();
+        printBuildTimeMetricsData(buildValidationData);
```
etiennestuder, line 78: My initial thought is that fetchBuildScanData should be a `Pair` rather than a `List`.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: `It's unclear if the Gradle Enterprise API could return these as null.` Did you check the OpenAPI definition?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * @return an estimation of the build time if all cacheable tasks had been
+     * avoided.
+     */
+    public Duration getPendingSavingsBuildTime() {
+        return pendingSavingsBuildTime;
+    }
+
+    private static Duration calculatePendingSavings(
```
etiennestuder, line 81: I don't think it's worth pulling out this simple calculation into its own method.

etiennestuder (changes requested):

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: https://docs.gradle.com/enterprise/api-manual/#reference_documentation

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
erichaagdev: This is [the same pattern used in the other "Fields" class](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/3870d12078a7410dceb439b53085ebed6cb63c98/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java#L48-L50). I kept it here for consistency. Would you like me to change it there too?

Though, I'm curious why you find it odd? There are many Java built-in methods that return a `Stream<..>` (e.g. `String#lines`). This simplifies the common use case of callers wanting to stream the enum values. Perhaps, the method could be named differently?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
erichaagdev: I did check it. We are already doing some `null` checking anyway on the response fields. It would be nice to isolate the `null` checking to one place or remove it completely.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
```diff
@@ -9,7 +9,7 @@
 import java.util.function.Function;
 import java.util.stream.Stream;
 
-public enum Fields {
+public enum BuildValidationDataFields {
```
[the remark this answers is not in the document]
erichaagdev: I named this class `BuildValidationDataFields` since it contains the fields of `BuildValidationData` which already existed.

This allowed me to follow the same convention for `BuildTimeMetrics` and `BuildTimeMetricsFields`.

I'm fine changing it, but we should probably also rename `BuildValidationData`.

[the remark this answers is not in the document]
erichaagdev: I updated `BuildValidationDataFields` -> `BuildScanDataFields` in 499b9e75414e49a17833b8cda17a0eb603efd821

The question is still open as to also rename `BuildValidationData` and `FetchBuildValidationDataCommand`.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+import java.math.BigDecimal;
+import java.time.Duration;
+
+public class BuildTimeMetrics {
+
+    private final Duration initialBuildTime;
+    private final Duration instantSavings;
+    private final Duration instantSavingsBuildTime;
+    private final Duration pendingSavings;
+    private final Duration pendingSavingsBuildTime;
```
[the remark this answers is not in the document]
erichaagdev: Done in 60c14402695b32c05290a7e03efe7c718334e0cb

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
erichaagdev: Removed in a4ce446a6a291ec80699ca65acc3400460e5336d

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,47 @@
...
+            Duration pendingSavings,
+            Duration pendingSavingsBuildTime) {
+        this.initialBuildTime = initialBuildTime;
+        this.instantSavings = instantSavings;
+        this.instantSavingsBuildTime = instantSavingsBuildTime;
+        this.pendingSavings = pendingSavings;
+        this.pendingSavingsBuildTime = pendingSavingsBuildTime;
+    }
+
+    public static BuildTimeMetrics from(BuildValidationData firstBuild, BuildValidationData secondBuild) {
```
erichaagdev, line 26: I feel this static constructor is too much. Should I condense this all to a single constructor that takes `BuildValidationData, BuildValidationData`?

components/scripts/lib/build-scan-online.sh
```diff
@@ -44,17 +44,53 @@ is_build_scan_metadata_missing() {
   return 0
 }
 
-fetch_and_read_build_scan_data() {
+fetch_and_read_single_build_scan() {
```
erichaagdev, line 47: I am not very happy with these function names. We have:

- `fetch_and_read_single_build_scan`
- `parse_single_build_scan`
- `fetch_and_read_build_scans`
- `parse_build_scans`

components/scripts/lib/build-scan-online.sh
```diff
@@ -44,17 +44,53 @@ is_build_scan_metadata_missing() {
...
+  local build_scan_url="$1"
+
+  local build_scan_data
+  build_scan_data="$(fetch_build_scan_data brief_logging "${build_scan_url}")"
+
+  debug "Raw build scan data"
+  debug "---------------------------"
+  debug "${build_scan_data}"
```
erichaagdev, line 55: You will see this `build_scan_data` nomenclature in my changes. I don't know if it's quite right because really it might contain more than Build Scan data. It can also contain the build time metrics. It may also possibly contain other data in the future.

components/scripts/lib/build-scan-online.sh
erichaagdev: If it helps, the use case for calling each one is clearly defined. 

`fetch_and_read_single_build_scan` is only called to get data to recreate a Build Scan (CI/Local experiments).

`fetch_and_read_build_scans` is called by all experiments so that the summary can be displayed.

components/scripts/lib/summary.sh
erichaagdev, line 1: `format_duration` is still hiding at the very end of this file. It is still needed because the build caching metrics durations are still returned as milliseconds.

I'm thinking we will also want those to be returned pre-formatted. I could update it now or in a subsequent PR. Let me know.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
etiennestuder: Let's keep it here for now.

I find a Stream not great here because it does not communicate anything about the data structure.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
etiennestuder: Got it.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
```diff
@@ -70,9 +71,12 @@ public Integer call() {
         logFinishedFetchingAllBuildScanData();
         logFetchResults(buildValidationData, customValueKeys);
 
-        printHeader();
+        printBuildValidationHeader();
```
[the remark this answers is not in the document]
etiennestuder: LGTM

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
etiennestuder: Ah, yes. CI/local will be different. But, IMO, printBuildTimeMetricsData should still take a Pair and not a List.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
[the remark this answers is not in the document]
etiennestuder: Nice.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: LGTM

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: Not sure how you would move everything to the constructor, since you also return null in some cases.

That said, I don't understand the need for these null checks. Non of these values should every be null as per the OpenApi spec.

components/scripts/lib/build-scan-online.sh
etiennestuder: Maybe something like this:

- fetch_single_build_scan
- parse_single_build_scan
- fetch_build_scans_and_build_time_metrics
- parse_build_scans_and_build_time_metrics

components/scripts/lib/summary.sh
etiennestuder: Let's do a separate PR for it.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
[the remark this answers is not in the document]
erichaagdev: Renamed in c32ab92b4ab9c5201a0c732cb3cf0164e1dbda93

components/scripts/lib/build-scan-online.sh
etiennestuder: One way out that I think would be easy to understand and distinguish is:
- _build scan data_ for everything that refers to data derived from a single build scan
- _build scan insights_ for everything that refers to an umbrella of build scan data and and additional metrics

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
erichaagdev: I agree. There are quite a few places we are using a `List` where a Pair or Tuple object would suffice.

Should that be done as part of this PR?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: Can't we just remove these null checks here? These values should never be null by contract.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
erichaagdev: > Not sure how you would move everything to the constructor, since you also return null in some cases.

Good point.

> That said, I don't understand the need for these null checks that then leads to a null object being returned. None of these values should every be null as per the OpenApi spec.

The `null` checks can be removed once we have an alternative [to returning a "default" object](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/c32ab92b4ab9c5201a0c732cb3cf0164e1dbda93/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java#L87-L112) for a failed Build Scan fetch.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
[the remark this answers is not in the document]
etiennestuder: LGTM, except to method renaming suggestions.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildValidationDataCommand.java
etiennestuder: Let's keep it as it is in this PR.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
etiennestuder: I see. Let's not do it in this PR.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
```diff
@@ -122,7 +122,7 @@ private Duration parseTimeout(String key) {
         }
     }
 
-    public BuildValidationData fetchBuildValidationData(NumberedBuildScan buildScan) {
+    public BuildScanData fetchBuildValidationData(NumberedBuildScan buildScan) {
```
etiennestuder, line 125: Rename method to `fetchBuildScanData`.

components/scripts/lib/build-scan-online.sh
erichaagdev: I like it. Updated in 0cb9dfd1e466d66c398b0493e93f8a5418c3b8c5.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
erichaagdev: > Can't we just remove these null checks here? These values should never be null by contract.

I think we came to a conclusion to this [in the other thread](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/383#discussion_r1152662319).

components/scripts/lib/build-scan-online.sh
```diff
@@ -7,7 +7,7 @@ readonly FETCH_BUILD_SCAN_DATA_JAR="${LIB_DIR}/export-api-clients/fetch-build-sc
 # Enterprise API.
 process_build_scan_data_online() {
   read_build_scan_metadata
-  fetch_and_read_build_scan_data build_cache_metrics_only
+  fetch_and_read_build_scans build_cache_metrics_only "${build_scan_urls[@]}"
```
etiennestuder: I keep forgetting why we need this `build_cache_metrics_only` flag and logic. Is it an over-optimization?

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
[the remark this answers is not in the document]
erichaagdev: Great catch. Fixed in d8d9203f6898a5b752c888e77339cbed23813389.

components/scripts/lib/build-scan-online.sh
erichaagdev: I really like this! It is exactly what I was looking for.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
erichaagdev: Done in d8d9203f6898a5b752c888e77339cbed23813389.

components/scripts/lib/build-scan-online.sh
erichaagdev: I had a similar thought while refactoring. It all boils down to this `if` statement in the parse logic: 

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/ecbb2ae2a328eb22c57bdf41724781b3a9f0e786/components/scripts/lib/build-scan-parse.sh#L64-L74

For CI/CI experiments the value is set to `all_data` because we don't have any information about the builds (e.g. outcome, requested tasks, etc.)

For CI/Local the **_first_** request to fetch the CI build is also `all_data` for a similar reason.

In all other cases it is set to `build_cache_metrics_only`. 

My thought was why not just parse all data all the time? We will override some values we already knew and obtained from the environment (e.g. requested tasks and build outcome). I think the only values the Build Scan may not have are the Git details in the scenario where CCUD is not applied. It would be nice to not have to pass the `build_cache_metrics_only` around.

etiennestuder: `My thought was why not just parse all data all the time? We will override some values we already knew and obtained from the environment (e.g. requested tasks and build outcome). I think the only values the Build Scan may not have are the Git details in the scenario where CCUD is not applied. It would be nice to not have to pass build_cache_metrics_only around to all these functions.`

Yes, we are fully aligned in our thoughts here. Can you add it to our doc, please, and then resolve the thread here?

components/scripts/lib/build-scan-online.sh
erichaagdev: I think I will hold off on doing this renaming for now. There are other occurrences of this `build_scan_data` nomenclature throughout the codebase, including in places not touched by this PR. There may even be some places we can apply this terminology to in the Java code too. 

I even think [these functions](https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/383#discussion_r1152658823) would be better renamed as: 

- `fetch_single_build_scan` -> `fetch_build_scan_data`
- `parse_single_build_scan` -> `parse_build_scan_data`
- `fetch_build_scans_and_build_time_metrics` -> `fetch_build_scan_insights`
- `parse_build_scans_and_build_time_metrics` -> `parse_build_scan_insights`

I will add this as a next step to discuss.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java
etiennestuder: LGTM

components/scripts/lib/build-scan-online.sh
etiennestuder: LGTM

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildValidationDataFields.java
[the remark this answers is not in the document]
etiennestuder: LGTM

components/scripts/lib/build-scan-online.sh
erichaagdev: Done.

components/scripts/lib/build-scan-online.sh
etiennestuder: SGTM

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
```diff
@@ -174,17 +174,43 @@ private void logFetchResultFor(int runNum, String property, String customValueKe
...
+
+        if (buildScanData.size() == 2) {
+            printBuildTimeMetricsHeader();
+            printBuildTimeMetrics(buildScanData);
+        }
+    }
+
+    public static void printBuildScanDataHeader() {
+        List<String> labels = BuildScanDataFields.ordered().map(f -> f.label).collect(Collectors.toList());
         System.out.println(String.join(",", labels));
```
jthurne, line 189: This could be simplified slightly by using [Collectors.joining](url):

```java
System.out.println(BuildScanDataFields.ordered().map(f -> f.label).collect(Collectors.joining(",")));
```

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
```diff
@@ -174,17 +174,43 @@ private void logFetchResultFor(int runNum, String property, String customValueKe
...
 
-    private void printBuildValidationData(List<BuildValidationData> buildValidationData) {
-        buildValidationData.forEach(this::printBuildValidationData);
+    private static void printBuildScanData(List<BuildScanData> buildScanData) {
+        buildScanData.forEach(FetchBuildScanDataCommand::printBuildScanData);
+    }
+
+    private static void printBuildScanData(BuildScanData buildScanData) {
+        List<String> values = BuildScanDataFields.ordered().map(f -> f.value.apply(buildScanData)).collect(Collectors.toList());
+        System.out.println(String.join(",", values));
```
jthurne, line 198: Similar simplification here:

```java
System.out.println(BuildScanDataFields.ordered().map(f -> f.value.apply(buildScanData)).collect(Collectors.joining(",")));
```

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
```diff
@@ -174,17 +174,43 @@ private void logFetchResultFor(int runNum, String property, String customValueKe
...
+    }
+
+    private static void printBuildTimeMetricsHeader() {
+        List<String> labels = BuildTimeMetricsFields.ordered().map(f -> f.label).collect(Collectors.toList());
+        System.out.println(String.join(",", labels));
```
jthurne, line 203:
```java
System.out.println(BuildTimeMetricsFields.ordered().map(f -> f.label).collect(Collectors.joining(",")));
```

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.time.Duration;
+
+public class BuildTimeMetrics {
+
+    public final Duration initialBuildTime;
+    public final Duration instantSavings;
+    public final Duration instantSavingsBuildTime;
+    public final Duration pendingSavings;
+    public final Duration pendingSavingsBuildTime;
```
jthurne, line 11: This class is now inconsistent with the other data classes. We should refactor the other data classes to follow this pattern as well.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/model/BuildTimeMetrics.java
```diff
@@ -0,0 +1,47 @@
...
+        this.pendingSavings = pendingSavings;
+        this.pendingSavingsBuildTime = pendingSavingsBuildTime;
+    }
+
+    public static BuildTimeMetrics from(BuildScanData firstBuild, BuildScanData secondBuild) {
+        if (firstBuild.getBuildTime() == null ||
+                secondBuild.getBuildTime() == null ||
+                secondBuild.getExecutedCacheableSummary() == null ||
+                secondBuild.getSerializationFactor() == null) {
+            return null;
```
jthurne, line 31: It might be better to return `Optional<BuildTimeMetrics>` rather than using `null`.

components/scripts/lib/build-scan-offline.sh
```diff
@@ -34,9 +34,9 @@ find_build_scan_dump() {
 }
 
 read_build_scan_dumps() {
-  local build_scan_csv
+  local build_scan_data
   echo -n "Extracting build scan data"
-  build_scan_csv="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" extract "0,${build_scan_dumps[0]}"  "1,${build_scan_dumps[1]}")"
-  parse_build_scan_csv "$build_scan_csv" "build_cache_metrics_only"
+  build_scan_data="$(invoke_java "$BUILD_SCAN_SUPPORT_TOOL_JAR" extract "0,${build_scan_dumps[0]}"  "1,${build_scan_dumps[1]}")"
+  parse_build_scan_data "build_cache_metrics_only" "$build_scan_data"
```
jthurne, line 40: We should drop the quotes around `build_cache_metrics_only`.  We don't quote this elsewhere.

components/scripts/lib/build-scan-parse.sh
```diff
@@ -25,38 +23,46 @@ executed_cacheable_duration=()
...
-    if [[ "$header_row_read" == "false" ]]; then
-      header_row_read=true
-      continue;
-    fi
+parse_build_scans_and_build_time_metrics() {
+  local build_cache_metrics_only="$1"
+  local build_scan_data="$2"
+
+  local build_scan_rows
+  IFS=$'\n' read -rd '' -a build_scan_rows <<< "$build_scan_data"
```
jthurne, line 50: This turned out rather nice.  I wonder if it makes sense to add a comment describing what this does (for those who are less familiar with bash).

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildTimeMetricsFields.java
```diff
@@ -0,0 +1,48 @@
...
```
jthurne, line 32: The only downside to doing this here is that we're now putting some of the display logic back into the Java code, where we used to have it only in the bash scripts.  I'm not saying we should undo this, just commenting that the separation of responsibilities is a bit muddy.

etiennestuder: Yes, the trend is and will be to have more in the Java code and less in the Bash scripts.

components/scripts/lib/build-scan-offline.sh
erichaagdev: I recognize the inconsistency but I intentionally did this. Without the quotes it is indistinguishable from a variable or function call. IMO it would be nice to add quotes in the other places.

components/scripts/lib/build-scan-parse.sh
erichaagdev: That's a good idea. I've already forgotten what it does. 🙃

components/scripts/lib/build-scan-offline.sh
jthurne: > it would be nice to add quotes in the other places.

That sounds reasonable to me. As long as how we call the function is consistent, I don't have a strong opinion here.

erichaagdev: I have addressed this in #388

components/scripts/lib/build-scan-parse.sh
erichaagdev: I have addressed this in #388

components/scripts/lib/build-scan-parse.sh
```diff
@@ -83,52 +89,20 @@ parse_build_scan_csv() {
...
+parse_build_time_metrics() {
+  local build_time_metrics_row="$1"
+
+  while IFS=, read -r field_1 field_2 field_3 field_4 field_5; do
+    initial_build_time="$field_1"
+    instant_savings="$field_2"
+    instant_savings_build_time="$field_3"
+    pending_savings="$field_4"
+    pending_savings_build_time="$field_5"
+  done <<< "${build_time_metrics_row}"
```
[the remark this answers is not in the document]
erichaagdev: Very nice! Addressed in #388

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
erichaagdev: Resolving as this was done in #391

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
erichaagdev: Resolving as this was done in #391

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java
erichaagdev: Resolving as this was done in #391

PR #385, opened 2023-03-30
New function name is used for offline processing

erichaagdev:
An occurrence of the old `parse_build_scan_data` function name was still referenced in the offline processing flow. This PR updates it to use the new function `parse_build_scans_and_build_time_metrics`.

This change also consolidates the debug logging for printing build validation data to `build-scan-parse.sh`, where it used to be. Otherwise the same 4 lines would have to be duplicated to the offline flow as well.


etiennestuder (approved):
LGTM

jthurne (approved):

PR #390, opened 2023-04-06
Serialization factor is formatted in Java

erichaagdev:
This PR moves the serialization rounding and formatting to the command line tool. 

image

The same rounding method used in GE is now used in the scripts. This provides a much more consistent experience.

image

image


components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildScanDataFields.java
```diff
@@ -81,4 +78,8 @@ private static String summaryTotal(BuildScanData data, String avoidanceOutcome,
     private static String formatDuration(Duration duration) {
         return duration == null ? NO_VALUE : format(duration);
     }
+
+    private static String formatSerializationFactor(BigDecimal serializationFactor) {
+        return serializationFactor == null ? NO_VALUE : serializationFactor.setScale(2, RoundingMode.HALF_UP).doubleValue() + "x";
```
erichaagdev, line 83: I was not sure if we also wanted to append the `x` here, rather than in the scripts. I went ahead and moved it so that more is being done in Java, but still wanted to ask.

etiennestuder: I don't feel strongly either way, thus moving to Java is ok.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildScanDataFields.java
```diff
@@ -81,4 +78,8 @@ private static String summaryTotal(BuildScanData data, String avoidanceOutcome,
     private static String formatDuration(Duration duration) {
         return duration == null ? NO_VALUE : format(duration);
     }
+
+    private static String formatSerializationFactor(BigDecimal serializationFactor) {
```
etiennestuder, line 82: Why is the serialization factor a BigDecimal?

etiennestuder (approved):
LGTM. Just one question to get clarity on. It does not block merging this to main.

components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/BuildScanDataFields.java
erichaagdev: Jim had the same comment in a previous PR. 

I've basically been conditioned to always default to using `BigDecimal` and to avoid `double` / `float` primitive types for representing floating point numbers [to avoid any chance of losing precision](https://stackoverflow.com/questions/588004/is-floating-point-math-broken).

I am fine to change it in a subsequent PR.

PR #396, opened 2023-04-12
Build Scan parsing is updated to read values when previously unknown

erichaagdev:
### Overview

This PR removes the `build_cache_metrics_only` and `all_data` values passed to and from various functions to control what summary data is parsed. The goal of this is to improve code readability by having one less variable to keep track of in the call stack.

The approach I decided to take was to set summary values only if they were empty. In the case of local experiments, a lot of the environmental summary values are already known. This approach prefers values determined by the scripts over those obtained from a Build Scan. Below is a summary of the changes:

- Previously, these fields were only set when `all_data` was passed to the parse function. Now, they are set when the current value is empty:
  - `base_urls`
  - `build_scan_urls`
  - `git_repos`
  - `git_branches`
  - `git_commit_ids`
  - `requested_tasks`
  - `build_outcomes`
- Previously, these fields were only set when `all_data` was passed to the parse function. Now they are always set:
  - `remote_build_cache_urls`
  - `remote_build_cache_shards`

### Testing

In order to test these changes, I compared the output before and after my changes. Below is an overview of the results. You will also see changes from #397 which is why the fetch logging is different.

<details>
<summary>Test results</summary>

#### Gradle - Experiment 3 (Before | After)

image

---

#### Gradle - Experiment 3 - No Gradle Enterprise API Access (Before | After)

image

---

#### Gradle - Experiment 4 (Before | After)

image

---

#### Gradle - Experiment 5 (Before | After)

image

---

#### Maven - Experiment 2 (Before | After)

image

---

#### Maven - Experiment 2 - No Gradle Enterprise API Access (Before | After)

image

---

#### Maven - Experiment 3 (Before | After)

image

---

#### Maven - Experiment 4 (Before | After)

image

</details>

### DIFF
--- a/components/scripts/lib/build-scan-online.sh
+++ b/components/scripts/lib/build-scan-online.sh
@@ -7,7 +7,7 @@ readonly FETCH_BUILD_SCAN_DATA_JAR="${LIB_DIR}/export-api-clients/fetch-build-sc
 # Enterprise API.
 process_build_scan_data_online() {
   read_build_scan_metadata
-  fetch_build_scans_and_build_time_metrics 'build_cache_metrics_only' "${build_scan_urls[@]}"
+  fetch_build_scans_and_build_time_metrics 'brief_logging' "${build_scan_urls[@]}"
 }
 
 read_build_scan_metadata() {
@@ -53,28 +53,27 @@ fetch_single_build_scan() {
   parse_single_build_scan "${build_scan_data}"
 }
 
+# The value of logging_level should be either 'brief_logging' or
+# 'verbose_logging'
 fetch_build_scans_and_build_time_metrics() {
-  local build_cache_metrics_only="$1"
+  local logging_level="$1"
   shift
   local build_scan_urls=("$@")
 
-  local brief_logging
-  if [[ "${build_cache_metrics_only}" == 'build_cache_metrics_only' ]]; then
-    brief_logging="brief_logging"
-  else
+  if [[ "${logging_level}" != 'brief_logging' ]]; then
     info "Fetching build scan data"
   fi
 
   local build_scan_data
-  build_scan_data="$(fetch_build_scan_data "${brief_logging}" "${build_scan_urls[@]}")"
+  build_scan_data="$(fetch_build_scan_data "${logging_level}" "${build_scan_urls[@]}")"
 
-  parse_build_scans_and_build_time_metrics "${build_cache_metrics_only}" "${build_scan_data}"
+  parse_build_scans_and_build_time_metrics "${build_scan_data}"
 }
 
 # Note: Callers of this function require stdout to be clean. No logging can be
 #       done inside this function.
 fetch_build_scan_data() {
-  local brief_logging="$1"
+  local logging_level="$1"
   shift
   local build_scan_urls=("$@")
 
@@ -90,7 +89,7 @@ fetch_build_scan_data() {
     args+=("--network-settings-file" "${SCRIPT_DIR}/network.settings")
   fi
 
-  if [[ "${brief_logging}" == "brief_logging" ]]; then
+  if [[ "${logging_level}" == "brief_logging" ]]; then
     args+=("--brief-logging")
   fi
 
--- a/components/scripts/lib/build-scan-online.sh
+++ b/components/scripts/lib/build-scan-online.sh
@@ -60,7 +60,7 @@ fetch_build_scans_and_build_time_metrics() {
   shift
   local build_scan_urls=("$@")
 
-  if [[ "${logging_level}" != 'brief_logging' ]]; then
+  if [[ "${logging_level}" == 'verbose_logging' ]]; then
     info "Fetching build scan data"
   fi
 
--- a/components/scripts/lib/build-scan-parse.sh
+++ b/components/scripts/lib/build-scan-parse.sh
@@ -43,12 +43,11 @@ parse_single_build_scan() {
   # Parses build scan data to an array by line
   IFS=$'\n' read -rd '' -a build_scan_rows <<< "$build_scan_data"
 
-  parse_build_scan_row 'all_data' "${build_scan_rows[1]}"
+  parse_build_scan_row "${build_scan_rows[1]}"
 }
 
 parse_build_scans_and_build_time_metrics() {
-  local build_cache_metrics_only="$1"
-  local build_scan_data="$2"
+  local build_scan_data="$1"
 
   debug_build_scan_data "$build_scan_data"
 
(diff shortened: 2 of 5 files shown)


etiennestuder:
@erichaagdev in the case of the CI/local experiment, I would have expected that we print more details about the CI build scan when we fetch it the first time (similar to CI/CI).

components/scripts/lib/build-scan-parse.sh
```diff
@@ -74,28 +73,48 @@ debug_build_scan_data() {
...
-      remote_build_cache_shards[run_num]="${field_11}"
     fi
 
+    if [ -z "${build_outcomes[run_num]}" ]; then
+      build_outcomes[run_num]="${field_9}"
+    fi
+
+    remote_build_cache_urls[run_num]="${field_10}"
+    remote_build_cache_shards[run_num]="${field_11}"
+
```
etiennestuder, line 117: What is field_20 (it seems not to be read further down) ?

components/scripts/lib/build-scan-parse.sh
```diff
@@ -74,28 +73,48 @@ debug_build_scan_data() {
...
-    build_scan_ids[run_num]="$field_4"
-
-    if [[ "$build_cache_metrics_only" != 'build_cache_metrics_only' ]]; then
-      base_urls[run_num]="$field_2"
-      build_scan_urls[run_num]="$field_3"
-      git_repos[run_num]="$field_5"
-      git_branches[run_num]="$field_6"
-      git_commit_ids[run_num]="$field_7"
+
+    project_names[run_num]="${field_1}"
```
etiennestuder, line 83: Should we do the same (I mean the inverse) for project_names? (for example, to cover the case where the export API cannot be accessed, it would be nice to fall back to the project names derived from the folder names)

etiennestuder:
> Successful cross platform tests run: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4674139330
> 
> @etiennestuder here you can see the full output for CI/Local: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4674139330/jobs/8278073687#step:11:12

Nice. LGTM

components/scripts/lib/build-scan-parse.sh
erichaagdev: Good question. Field 20 is "Build time". It is no longer used as we return "Initial build time" and "Instant build time" in the new build time metrics part of the response. It probably should have been removed as part of #383.

etiennestuder: I see. Thanks.

components/scripts/lib/build-scan-parse.sh
erichaagdev: I had made a change in #263 to always resolve the project name from the Gradle Enterprise API, and removed code relating to reading project names from Gradle and Maven. 

It seems that even after being reviewed and even after all the testing I did on the PR, the case where Gradle Enterprise API access is not available was not considered

I just tested it, and if API access is not configured, the project name will be unknown. 

image

Do we want to add this functionality back? If so, then your comment makes sense to me.

etiennestuder: I suggest we do add it back as a fallback in case the EAPI cannot be called. But let's do that in a separate PR.

components/scripts/lib/build-scan-online.sh
```diff
@@ -53,28 +53,27 @@ fetch_single_build_scan() {
...
-  local build_cache_metrics_only="$1"
+  local logging_level="$1"
   shift
   local build_scan_urls=("$@")
 
-  local brief_logging
-  if [[ "${build_cache_metrics_only}" == 'build_cache_metrics_only' ]]; then
-    brief_logging="brief_logging"
-  else
+  if [[ "${logging_level}" != 'brief_logging' ]]; then
```
jthurne, line 63: 🧐 You could reverse this to be a positive test for "verbose_logging" instead. Positive tests tend to be easier to understand.

components/scripts/lib/build-scan-online.sh
```diff
@@ -53,28 +53,27 @@ fetch_single_build_scan() {
   parse_single_build_scan "${build_scan_data}"
 }
 
+# The value of logging_level should be either 'brief_logging' or
+# 'verbose_logging'
```
jthurne, line 57: 💡In the future, we could rename these to just "brief" and "verbose". Including "logging" here is arguably redundant.   We could also update the fetch tool to support this with an actual CLI parameter rather than reading it from a positional parameter.

Thinking about it, "quiet" might be a better term. "quiet" is used more often than "brief" in other applications for this sort of thing.

components/scripts/lib/build-scan-parse.sh
```diff
@@ -74,28 +73,48 @@ debug_build_scan_data() {
 
 # shellcheck disable=SC2034 # not all scripts use all of the fetched data
 parse_build_scan_row() {
-  local build_cache_metrics_only="$1"
-  local build_scan_row="$2"
+  local build_scan_row="$1"
 
   local run_num
 
   while IFS=, read -r run_num field_1 field_2 field_3 field_4 field_5 field_6 field_7 field_8 field_9 field_10 field_11 field_12 field_13 field_14 field_15 field_16 field_17 field_18 field_19 field_20 field_21; do
```
jthurne, line 80: 💡The more we work with this code, the more it bugs me that these are all called "field_N". At some point, we really should try to rename them.

jthurne (approved):
All of my comments are more about ideas for further improvements. The change itself looks good.

components/scripts/lib/build-scan-online.sh
erichaagdev: I actually did consider going with just `brief` and `verbose` but I was afraid it would not be clear enough about what it meant. If you just see `verbose` at the call site, it's not clear what is supposed to be verbose without digging into the function's code. For example:

```bash
fetch_build_scans_and_build_time_metrics 'verbose' "${build_scan_urls[@]}"
```

Then again, it is common for command line tools to have `--quiet` and `--verbose` flags and it is usually clear what it means. So maybe I am overthinking it. I would be fine to change it to how you describe if you think that it is clear enough. 

> We could also update the fetch tool to support this with an actual CLI parameter rather than reading it from a positional parameter

Unless I misunderstand what you mean, this is already the case.

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/97caa96500593310919013e34df9f61e30523df5/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/cli/FetchBuildScanDataCommand.java#L52-L53

> Thinking about it, "quiet" might be a better term. "quiet" is used more often than "brief" in other applications for this sort of thing.

Ha, I do think it is more common, but usually quiet means no output. In this case we still output some details, just not as much.

components/scripts/lib/build-scan-online.sh
erichaagdev: That is a good point. Done in 2428dacd36f1f86a4c9f8ca66a2f36c1efa05037.

erichaagdev:
@jthurne I went forward with the one suggestion you made. If we want to rename the `brief` and `verbose` parameters I can do that in a subsequent PR. 

Merging this one as it has been approved.

components/scripts/lib/build-scan-online.sh
etiennestuder: I would stick to the current values. They are a bit more expressive.

PR #397, opened 2023-04-12
Logging level when fetching Build Scan data from Gradle Enterprise is incorrect

erichaagdev:
This change fixes a regression introduced by #383 where the conditions under which `--brief-logging` were incorrectly switched.

**Behavior in v2.3.2 (correct)**

image

**Behavior in v2.3.3 (incorrect)**

image

**Behavior with these changes (correct)**

image

### DIFF
--- a/components/scripts/lib/build-scan-online.sh
+++ b/components/scripts/lib/build-scan-online.sh
@@ -48,7 +48,7 @@ fetch_single_build_scan() {
   local build_scan_url="$1"
 
   local build_scan_data
-  build_scan_data="$(fetch_build_scan_data verbose_logging "${build_scan_url}")"
+  build_scan_data="$(fetch_build_scan_data 'verbose_logging' "${build_scan_url}")"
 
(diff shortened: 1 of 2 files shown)


components/scripts/lib/build-scan-online.sh
```diff
@@ -48,7 +48,7 @@ fetch_single_build_scan() {
   local build_scan_url="$1"
 
   local build_scan_data
-  build_scan_data="$(fetch_build_scan_data brief_logging "${build_scan_url}")"
+  build_scan_data="$(fetch_build_scan_data verbose_logging "${build_scan_url}")"
```
etiennestuder, line 51: Should we use quotes for 'verbose_logging' for consistency? Not sure, just asking.

erichaagdev: Yes, we should. Good catch. Addressed in 98bef4add67aac7d6bc04eb223b41fffffb22b9d

PR #399, opened 2023-04-12
Use Provider API to read version.txt and changes.md files

tylerbertrand:
* Add release notes to githubRelease body
* Use Provider.fileContents() to read `changes.md` and `version.txt` files

### Tests

Execute `./gradlew githubRelease` with `dryRun.set(true)`

```
:githubRelease [This task is a dry run. All API calls that would modify the repo are disabled. API calls that access the repo information are not disabled. Use this to show what actions would be executed.]
:githubRelease [CHECKING FOR PREVIOUS RELEASE]
:githubRelease [EXISTING RELEASE FOUND https://github.com/gradle/gradle-enterprise-build-validation-scripts/releases/tag/development-latest]
:githubRelease [DELETING RELEASE Development Build]
:githubRelease [CREATING NEW RELEASE 
{
    tag_name               = development-latest
    target_commitish       = main
    name                   = Development release
    generate_release_notes = false
    body                   = 
        - [FIX] Logging level when fetching Build Scan data from Gradle Enterprise is incorrect 
                - [FIX] Experiments accepting Build Scan URLs incorrectly parse Build Scan ID when path contains extra parts
    draft                  = false
    prerelease             = true
}]
```

And with `finalRelease` property set...

```
:githubRelease [This task is a dry run. All API calls that would modify the repo are disabled. API calls that access the repo information are not disabled. Use this to show what actions would be executed.]
:githubRelease [CHECKING FOR PREVIOUS RELEASE]
:githubRelease [CREATING NEW RELEASE 
{
    tag_name               = v2.3.4
    target_commitish       = main
    name                   = 2.3.4
    generate_release_notes = false
    body                   = 
        - [FIX] Logging level when fetching Build Scan data from Gradle Enterprise is incorrect 
                - [FIX] Experiments accepting Build Scan URLs incorrectly parse Build Scan ID when path contains extra parts
    draft                  = false
    prerelease             = false
}]
```

### Configuration Cache Compatibility

Modifying `version.txt` results in a new configuration cache entry being stored, and the dry run `githubRelease` uses the updated version.

Modifying `changes.md` results in a new configuration cache entry being stored, and the dry run `githubRelease` uses the updated body.

Running with no changes results in the configuration cache entry being re-used.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,9 +29,11 @@ repositories {
     mavenCentral()
 }
 
-val appVersion = layout.projectDirectory.file("release/version.txt").asFile.readText().trim()
+val releaseVersion = releaseVersion()
+val releaseNotes = releaseNotes()
+
 allprojects {
-    version = appVersion
+    version = releaseVersion
 }
 
 val argbash by configurations.creating
@@ -93,8 +95,10 @@ val copyGradleScripts by tasks.registering(Copy::class) {
     group = "build"
     description = "Copies the Gradle source and generated scripts to output directory."
 
-    val projectVersion = appVersion // local variable required for configuration cache compatibility
-    inputs.property("project.version", projectVersion)
+    // local variable required for configuration cache compatibility
+    // https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:not_yet_implemented:accessing_top_level_at_execution
+    val releaseVersion = releaseVersion
+    inputs.property("project.version", releaseVersion)
 
     from(layout.projectDirectory.file("LICENSE"))
     from(layout.projectDirectory.dir("release").file("version.txt"))
@@ -102,7 +106,7 @@ val copyGradleScripts by tasks.registering(Copy::class) {
 
     from(layout.projectDirectory.dir("components/scripts/gradle")) {
         exclude("gradle-init-scripts")
-        filter { line: String -> line.replace("<HEAD>", projectVersion) }
+        filter { line: String -> line.replace("<HEAD>", releaseVersion.get()) }
     }
     from(layout.projectDirectory.dir("components/scripts/gradle")) {
         include("gradle-init-scripts/**")
@@ -114,7 +118,7 @@ val copyGradleScripts by tasks.registering(Copy::class) {
         include("network.settings")
         include("lib/**")
         exclude("lib/cli-parsers")
-        filter { line: String -> line.replace("<HEAD>", projectVersion) }
+        filter { line: String -> line.replace("<HEAD>", releaseVersion.get()) }
     }
     from(generateBashCliParsers.map { it.outputDir.file("lib/cli-parsers/gradle") }) {
         into("lib/")
@@ -129,15 +133,17 @@ val copyMavenScripts by tasks.registering(Copy::class) {
     group = "build"
     description = "Copies the Maven source and generated scripts to output directory."
 
-    val projectVersion = appVersion // local variable required for configuration cache compatibility
-    inputs.property("project.version", projectVersion)
+    // local variable required for configuration cache compatibility
+    // https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:not_yet_implemented:accessing_top_level_at_execution
+    val releaseVersion = releaseVersion
+    inputs.property("project.version", releaseVersion)
 
     from(layout.projectDirectory.file("LICENSE"))
     from(layout.projectDirectory.dir("release").file("version.txt"))
     rename("version.txt", "VERSION")
 
     from(layout.projectDirectory.dir("components/scripts/maven")) {
-        filter { line: String -> line.replace("<HEAD>", projectVersion) }
+        filter { line: String -> line.replace("<HEAD>", releaseVersion.get()) }
     }
     from(layout.projectDirectory.dir("components/scripts/")) {
         include("README.md")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -253,7 +253,8 @@ githubRelease {
     prerelease.set(isDevelopmentRelease)
     overwrite.set(isDevelopmentRelease)
     generateReleaseNotes.set(false)
-    body.set(layout.projectDirectory.file("release/changes.md").asFile.readText().trim())
+    body.set(releaseNotes())
+    dryRun.set(true) // REMOVE BEFORE MERGE
```
etiennestuder, line 257: Please resolve this comment once you have removed the line.

build.gradle.kts
```diff
@@ -272,26 +273,36 @@ tasks.withType(Sign::class).configureEach {
...
-    if (isDevelopmentRelease) {
-        return "Development Build"
+fun gitHubReleaseName(): Provider<String> {
+    return if (isDevelopmentRelease) {
+        provider { "Development Build" }
     } else {
-        return version.toString()
+        appVersion
     }
 }
```
etiennestuder, line 282: I _think_ this could be shortened to `appVersion.map { isDevelopmentRelease ? "Development Build" : it}`
Similarly for the other two methods gitReleaseTag and distributionVersion.

Please give it a try and also verify CC-compatibility.

etiennestuder (commented):
Thanks, Tyler.

build.gradle.kts
tylerbertrand: Is one of these the preferred style?

tylerbertrand: and would there be any reason not to apply that to the other similar functions?
```
fun gitHubReleaseName() = appVersion.map { if (isDevelopmentRelease) "Development release" else it }

fun gitReleaseTag() = appVersion.map { if (isDevelopmentRelease) "development-latest" else "v$it" }

fun distributionVersion() = appVersion.map { if (isDevelopmentRelease) "dev" else it}
```

erichaagdev: For small functions, the expression body (the second one) is more idiomatic Kotlin.

But, ultimately it is personal preference. Sometimes I will choose the former for consistency because its the only function that could be turned use an expression body.

I will leave it between you and @etiennestuder.

tylerbertrand: Just pushed a commit with the code snippet I posted above

etiennestuder: > Is one of these the preferred style?

I would have chosen the style that declares the return type and is also what the other methods use. I know it might not be as concise as it can be, so in the end, it is just my personal preference for expressiveness over conciseness.

```
fun xyz: Provider<String> {
```

tylerbertrand: > Similarly for the other two methods gitReleaseTag and distributionVersion.

Whoops, must have glanced over that, sorry.

> `fun xyz: Provider<String> {`

Yeah I agree, that style is clearer. Not having much Kotlin experience, I just thought to go with what might be considered the more idiomatic approach. I'll update to use this style, as it's both and clearer and more consistent with the other functions.

etiennestuder: Thanks, Tyler.

etiennestuder: @tylerbertrand Feel free to merge once you and @erichaagdev are happy with it.

tylerbertrand: Sounds good - thanks @etiennestuder!

build.gradle.kts
```diff
@@ -253,7 +253,8 @@ githubRelease {
     prerelease.set(isDevelopmentRelease)
     overwrite.set(isDevelopmentRelease)
     generateReleaseNotes.set(false)
-    body.set(layout.projectDirectory.file("release/changes.md").asFile.readText().trim())
+    body.set(releaseNotes())
```
erichaagdev, line 256: IIRC in the other PRs you added a `releaseNotes` variable at the top of the script.. I would do the same so that everything is consistent.

build.gradle.kts
```diff
@@ -29,7 +29,7 @@ repositories {
     mavenCentral()
 }
 
-val appVersion = layout.projectDirectory.file("release/version.txt").asFile.readText().trim()
+val appVersion = releaseVersion()
```
erichaagdev, line 32: We have 4 different version variables in this script that _I think_ are all the same thing. 

`appVersion`, `releaseVersion`, `projectVersion`, and the normal `version`. 😵 

In the other scripts we just have `releaseVersion` which gets assigned to `version`.

Is it possible to do the same thing here? I realize you are just following the same patterns that already existed in this build script, but with the addition of the `releaseVersion()` function it might be a good time to consider cleaning it up.

build.gradle.kts
tylerbertrand: Can do. The reason I did not do that here is that it is only used in one place (to set the `githubRelease` body).

build.gradle.kts
tylerbertrand: I can investigate - `projectVersion` has a comment saying it is required for configuration cache compatibility. I wasn't exactly sure why that was the case, but I'll look into that and into consolidating those variables.

tylerbertrand: After looking into this, unless you know a way around it, the `projectVersion` variable is unavoidable (though maybe it could use a better name).

From the docs [here](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:not_yet_implemented:accessing_top_level_at_execution)...
```
Build scripts written in Kotlin cannot store tasks that reference top-level methods or variables at execution time in the configuration cache at all. This limitation exists because the captured script object references cannot be serialized.
```

erichaagdev: Tyler and I paired on this to come up with a solution. Now there is only `releaseVersion` and the regular `project.version` which is just like our other repositories.

erichaagdev (approved):

PR #400, opened 2023-04-13
Project names are displayed in summary when access to Gradle Enterprise API is not available

erichaagdev:
In #263 a change was made to always read the project name from the Gradle Enterprise API. All code around extracting the project name locally was removed. Unfortunately, the use case of not having access to the Gradle Enterprise API was not considered and therefore a regression was introduced. 

Currently, without these changes, if Gradle Enterprise API access is not configured, the project name will be unknown. 

image

With these changes, the project name will be shown even without Gradle Enterprise API access.

image

And for Maven (before | after):

image

### DIFF
--- a/components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java
+++ b/components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java
@@ -1,0 +1,73 @@
+package com.gradle;
+
+import org.apache.maven.execution.ExecutionEvent;
+import org.apache.maven.execution.MavenExecutionRequest;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.building.ModelProcessor;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.project.ProjectBuilder;
+import org.apache.maven.project.ProjectBuildingException;
+import org.codehaus.plexus.logging.Logger;
+
+import javax.inject.Inject;
+import java.io.File;
+import java.util.*;
+
+import static java.util.Collections.emptyList;
+
+public class RootProjectExtractor {
+    private final ProjectBuilder projectBuilder;
+    private final ModelProcessor modelProcessor;
+    private final Logger logger;
+
+    @Inject
+    public RootProjectExtractor(ProjectBuilder projectBuilder, ModelProcessor modelProcessor, Logger logger) {
+        this.projectBuilder = projectBuilder;
+        this.modelProcessor = modelProcessor;
+        this.logger = logger;
+    }
+
+    public MavenProject extractRootProject(ExecutionEvent event) {
+        return extractRootProject(event.getSession());
+    }
+
+    public MavenProject extractRootProject(MavenSession session) {
+        List<MavenProject> allProjects = discoverAllProjects(session.getAllProjects());
+        File workspaceDirectory = getWorkspaceDirectory(session);
+
+        if (workspaceDirectory.equals(allProjects.get(0).getBasedir())) {
+            return allProjects.get(0);
+        }
+
+        File workspaceDirectoryPom = modelProcessor.locatePom(workspaceDirectory);
+        if (workspaceDirectoryPom.exists()) {
+            try {
+                return projectBuilder.build(workspaceDirectoryPom, session.getProjectBuildingRequest()).getProject();
+            } catch (ProjectBuildingException e) {
+                logger.error("Exception locating the top level project", e);
+            }
+        }
+
+        // We didn't successfully identify the root project, so just return the first project.
+        return allProjects.get(0);
+    }
+
+    private File getWorkspaceDirectory(MavenSession session) {
+        MavenExecutionRequest request = session.getRequest();
+        try {
+            return request.getMultiModuleProjectDirectory();
+        } catch (NoSuchMethodError ignored) {
+            return new File(session.getExecutionRootDirectory());
+        }
+    }
+
+    /**
+     * Older Maven versions under-reported the list of discovered projects.
+     * This method discovers all their submodules. For newer Maven versions it is a no-op.
+     */
+    private List<MavenProject> discoverAllProjects(Collection<MavenProject> sessionProjects) {
+        Set<MavenProject> allProjects = new LinkedHashSet<>(sessionProjects);
+        sessionProjects.stream().flatMap(p -> Optional.ofNullable(p.getCollectedProjects()).orElse(emptyList()).stream()).forEach(allProjects::add);
+        return new ArrayList<MavenProject>(allProjects);
+    }
+}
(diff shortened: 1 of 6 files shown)


components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java
erichaagdev, line 1: This is an exact copy of [the `RootProjectExtractor` we had previously](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/da175ff75c94e3807ced83167ed779e2dd2d8677/components/capture-build-scan-url-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java). I did not attempt to modify it in any way.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
```diff
@@ -105,13 +105,14 @@ def registerBuildScanActions = { def buildScan ->
             return System.getProperty(name) ?: System.getenv(envVarName)
         }
         def runNum = getRunNumInputParam('com.gradle.enterprise.build-validation.runNum')
+        def rootProjectName = gradle.rootProject.name
```
erichaagdev, line 108: Previously, the root project name was extracted using `settings.rootProject.name` for Gradle versions >= 6, or from `gradle.rootProject.name` for Gradle versions >= 5. See below.

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/da175ff75c94e3807ced83167ed779e2dd2d8677/components/scripts/gradle/gradle-init-scripts/capture-published-build-scan.gradle#L15

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/da175ff75c94e3807ced83167ed779e2dd2d8677/components/scripts/gradle/gradle-init-scripts/capture-published-build-scan.gradle#L22

I think it was probably due to the type of hook being used. In my testing, extracting the root project name via `gradle.rootProject.name` worked in all Gradle versions supported by the scripts (5+), so I did not think it made sense to try and extract it differently based on Gradle version.

components/scripts/lib/build-scan-online.sh
```diff
@@ -24,10 +24,14 @@ read_build_scan_metadata() {
       debug ""
     fi
 
-    while IFS=, read -r run_num field_1 field_2 field_3; do
-       base_urls[$run_num]="$field_1"
-       build_scan_urls[$run_num]="$field_2"
-       build_scan_ids[$run_num]="$field_3"
+    local run_num project_name base_url build_scan_url build_scan_id
```
erichaagdev, line 27: @jthurne Is this sort of what you had in mind for updating the Build Scan parsing logic? Since I was modifying the `build-scans.csv` code I figured I could at least clean this up a little bit.

components/scripts/lib/build-scan-parse.sh
```diff
@@ -112,6 +120,10 @@ parse_build_scan_row() {
       build_outcomes[run_num]="${field_9}"
     fi
 
+    # The below fields are always set by Build Scan data regardless of their
+    # previous value and are always safe to override.
+
+    build_scan_ids[run_num]="${field_4}"
```
erichaagdev, line 126: Even though this is field 4, I thought it made sense to move it down here so that all fields are grouped according to how they are set. Let me know what you think.

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java
etiennestuder: 👍🏻

components/scripts/lib/build-scan-online.sh
etiennestuder: I'm not Jim :-) but it reads nicely since one can now understand what the expected fields are.

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/RootProjectExtractor.java
```diff
@@ -0,0 +1,73 @@
...
+    private final Logger logger;
+
+    @Inject
+    public RootProjectExtractor(ProjectBuilder projectBuilder, ModelProcessor modelProcessor, Logger logger) {
+        this.projectBuilder = projectBuilder;
+        this.modelProcessor = modelProcessor;
+        this.logger = logger;
+    }
+
+    public MavenProject extractRootProject(ExecutionEvent event) {
```
etiennestuder: This method is unused. I believe it can be removed.

etiennestuder: Done.

etiennestuder (approved):

etiennestuder (changes requested):
One suggestion open.

components/scripts/lib/build-scan-parse.sh
erichaagdev: I think what you are saying is that if we already know what the Build Scan ID is then there is no reason to override it?

That is actually the same way we treat `base_urls` and `build_scan_urls` so actually I think it can move to be right next to those.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
erichaagdev: I just discovered accessing `gradle` from a Groovy closure is not supported by the configuration cache. I will have to rethink this a little.

erichaagdev: See here: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4692331201/jobs/8317949102#step:8:77

(For some reason it is not captured in the Build Scan)

erichaagdev: To solve this, I added a `rootProjectName` parameter to the `registerBuildScanActions` function and call either `settings.rootProject.name` or `rootProject.name` depending on Gradle version. 

Both the cross Gradle version tests I ran locally and the [cross platform tests](https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/4692651896/jobs/8318658237) seemed to like it.

components/scripts/lib/build-scan-online.sh
etiennestuder: I'll close this thread.

components/scripts/lib/build-scan-parse.sh
etiennestuder: LGTM.

components/scripts/gradle/gradle-init-scripts/configure-gradle-enterprise.gradle
etiennestuder: Very nice.

Interestingly, it is no also very close to how we pass the root project name in the case of Maven.

etiennestuder (approved):
LGTM

PR #409, opened 2023-04-17
Unused build time field is removed from Java command line tool response

erichaagdev:
In #383, the build time field was superseded by the "initial" and "instant" build time fields. This PR removes all references to the build time field as it is no longer used.


etiennestuder (approved):
LGTM. Same change to be done in the offline mode?

erichaagdev:
> LGTM. Same change to be done in the offline mode?

Yep. It's finished, just testing it. Will have a PR shortly.

PR #411, opened 2023-04-18
Performance characteristics are explained in README and link included in experiment summary

erichaagdev:
image

### DIFF
--- a/Gradle.md
+++ b/Gradle.md
@@ -161,6 +161,18 @@ The summary looks typically like in the screenshot below.
 
 ![image](https://user-images.githubusercontent.com/231070/146644509-7b071436-3d04-46ee-a327-ea16ecfe9c76.png)
 
+### Performance characteristics
+
+| Characteristic                  | Description |
+|---------------------------------|-------------|
+| Initial build time              |             |
+| Build time with instant savings |             |
+| Build time with pending savings |             |
+| Avoided cacheable tasks         |             |
+| Executed cacheable tasks        |             |
+| Executed non-cacheable tasks    |             |
+| Serialization factor            |             |
+
 ## Investigating file resources on the local file system
 
 For the scripts that run one or more builds locally, the file resources that are used and produced by the builds can be investigated on the local file system. This is helpful when trying to understand cache misses due to changes in file inputs of the executed tasks. All intermediate and final output produced while running a given script is stored under `.data/<script_name>/<timestamp>-<run_id>`.
--- a/Gradle.md
+++ b/Gradle.md
@@ -177,6 +177,7 @@ The elapsed time of the second build with build performance acceleration measure
 Achieving the build time with instant savings does not require any changes to the tasks of the build.
 
 #### Build time with pending savings
+
 The projected elapsed build time of the second build with build performance acceleration measures explicitly applied by the experiment. The projection assumes that all cacheability issues of the executed cacheable tasks get resolved, and it takes into account the degree of parallelization in task execution.
 
 Achieving the build time with pending savings requires changes to the executed cacheable tasks of the build.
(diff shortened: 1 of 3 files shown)


Maven.md
```diff
@@ -159,6 +159,46 @@ The summary looks typically like in the screenshot below.
...
+
+Achieving the build time with pending savings requires changes to the executed cacheable goals of the build.
+
+#### Avoided cacheable goals
+
+The estimated reduction in serial execution time of the goals of the second build due to reusing the goals outputs from the first build. The avoidance savings are calculated as the difference between the time required to reuse the goals outputs and the time it took to create the goals outputs originally.
+
+#### Executed cacheable goals
+
+The serial execution time of the goals executed in the second build that Gradle considered cacheable. These goals stored their outputs in the build cache during the first build but were unable to reuse the outputs during the second build.
```
erichaagdev, line 188: > that Gradle considered cacheable

This sounds a little strange in the Maven README. What do you think about this:

> that Gradle Enterprise considered cacheable

erichaagdev: Same below in "Executed non-cacheable goals".

components/scripts/lib/summary.sh
```diff
@@ -181,10 +181,19 @@ print_performance_characteristics() {
...
   print_bl
-  info "Performance Characteristics"
+  info "Performance Characteristics ($(analyzing_the_results_link))"
   info "---------------------------"
 }
 
+# TODO update these links
+analyzing_the_results_link() {
+  if [[ "${BUILD_TOOL}" == "Maven" ]]; then
+    echo "https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/erichaagdev/performance-characteristics-explained/Maven.md#performance-characteristics"
```
erichaagdev, line 191: todo: update these links

It would be a nice improvement if these links did not always point to `main`, but to the tag for released versions. However, that is not something we do anywhere else in the code. 

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/e6eae99d02a18a91dcc5a0b01c98b28cf30978a0/components/scripts/lib/interactive-mode.sh#L48

If we wanted to do that, it should probably be a future enhancement (subsequent PR).

Gradle.md
```diff
@@ -161,6 +161,46 @@ The summary looks typically like in the screenshot below.
...
+
+The elapsed time of the first build without any build performance acceleration measures explicitly applied by the experiment.
+
+#### Build time with instant savings
+
+The elapsed time of the second build with build performance acceleration measures explicitly applied by the experiment. The experienced build performance acceleration typically comes from the incremental build feature (experiment 1) or build caching (experiments 2 to 5).
+
+Achieving the build time with instant savings does not require any changes to the tasks of the build.
+
+#### Build time with pending savings
```
erichaagdev, line 179: We should consistently add a newline after each heading.

Maven.md
etiennestuder: Agreed. Fixed in https://github.com/gradle/gradle-enterprise-build-validation-scripts/commit/938475e1a6cec5b541e755764f66b805e056b3af

Gradle.md
etiennestuder: Fixed in https://github.com/gradle/gradle-enterprise-build-validation-scripts/commit/938475e1a6cec5b541e755764f66b805e056b3af

components/scripts/lib/summary.sh
etiennestuder: Yes, I would not do this for now. It's not something we don't do for our GE plugin/extension, neither.
But, you are right in that it would be an alternative.

etiennestuder (approved):
LGTM.

PR #415, opened 2023-04-19
Built scripts are exported as consumable build artifact

erichaagdev:
This PR extends the build so that the scripts are exported as a consumable build artifact. This makes it so that the built scripts can be included and referenced in a composite build.


build.gradle.kts
```diff
@@ -9,6 +11,8 @@ plugins {
     id("org.gradle.wrapper-upgrade") version "0.11.1"
 }
 
+group = "org.gradle"
```
etiennestuder, line 14: Please use `com.gradle`, like we do for CCUD, etc. (given this is GE related).

build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("HasPlatformType")
```
etiennestuder, line 1: Do we need this and if so, why?

build.gradle.kts
```diff
@@ -171,6 +175,23 @@ tasks.register<Task>("copyScripts") {
     dependsOn(copyGradleScripts, copyMavenScripts)
 }
 
+val consumableGradleScripts by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+    attributes.attribute(Usage.USAGE_ATTRIBUTE, objects.named("gradle-build-validation-scripts"))
```
etiennestuder, line 181: I'm not familiar with the Usage.USAGE_ATTRIBUTE attribute? Can you add a pointer to the review, just for my education, please?

etiennestuder (changes requested):

build.gradle.kts
erichaagdev: [Here are the docs](https://docs.gradle.org/current/userguide/variant_attributes.html#ecosystem_independent_standard_attributes) and [here is the Javadoc](https://docs.gradle.org/current/javadoc/org/gradle/api/attributes/Usage.html)..

build.gradle.kts
erichaagdev: It is to suppress these warnings:

image

Alternatively, we could specify the type explicitly. For example:

```kotlin
val consumableGradleScripts: Configuration by configurations.creating {
  ...
}
```

build.gradle.kts
erichaagdev: Done in 20e1a85d9309aa7e2a9028d1dd53c7246611bb90

build.gradle.kts
etiennestuder: Thanks for the insights.

build.gradle.kts
etiennestuder: Thanks.

etiennestuder (approved):

Issue #429, opened 2023-05-08
`experiment-config-cache` action doesn't appear to grep correctly


ZacSweers:
When trying to set up a job on github actions with the `experiment-config-cache` action, the log appears to show a successful build but the job fails and exits with `3` anyway.

The log
```
Starting a Gradle Daemon, 1 stopped Daemon could not be reused, use --status for details
Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: :libraries:foundation:json:test
Type-safe project accessors is an incubating feature.

> Configure project :
The Dependency Analysis plugin is only known to work with versions of AGP between 4.2.2 and 7.4.2. You are using 8.0.1. Proceed at your own risk.

> Task :libraries:foundation:timber:processResources NO-SOURCE
> Task :libraries:foundation:timber:test:processResources NO-SOURCE
> Task :libraries:foundation:json:processTestResources NO-SOURCE
> Task :libraries:foundation:slack-commons:processResources NO-SOURCE
> Task :libraries:foundation:test-commons:processResources NO-SOURCE
> Task :libraries:foundation:json:processResources NO-SOURCE
> Task :libraries:foundation:jsr305:processResources NO-SOURCE
> Task :libraries:foundation:jsr305:compileKotlin FROM-CACHE
> Task :libraries:foundation:jsr305:compileJava NO-SOURCE
> Task :libraries:foundation:jsr305:classes UP-TO-DATE
> Task :libraries:foundation:timber:compileKotlin FROM-CACHE
> Task :libraries:foundation:timber:compileJava NO-SOURCE
> Task :libraries:foundation:timber:classes UP-TO-DATE
> Task :libraries:foundation:test-commons:compileKotlin FROM-CACHE
> Task :libraries:foundation:test-commons:compileJava NO-SOURCE
> Task :libraries:foundation:test-commons:classes UP-TO-DATE
> Task :libraries:foundation:jsr305:jar
> Task :libraries:foundation:timber:jar
> Task :libraries:foundation:test-commons:jar
> Task :libraries:foundation:timber:test:compileKotlin FROM-CACHE
> Task :libraries:foundation:timber:test:compileJava NO-SOURCE
> Task :libraries:foundation:timber:test:classes UP-TO-DATE
> Task :libraries:foundation:timber:test:jar
> Task :libraries:foundation:slack-commons:compileKotlin FROM-CACHE
[...]

PR #431, opened 2023-05-09
Allow configuration cache assertions to support multiple build runs

jprinet:
The experiment-config-cache GH action is not supporting multiple build experiment runs as we are looking explicitly for [one single execution](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/9f93fa0ea247622b2e2375f322647502fd8e34bd/.github/actions/gradle/experiment-config-cache/action.yml#L31).

This PR fixes https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/429


runningcode (approved):
Thanks for fixing this so quickly!

erichaagdev (approved):

PR #441, opened 2023-05-19
API requests allow some time for build scans to become available

erichaagdev:
This is to help address issues caused by recently published build scans not yet being ready for consumption from the Gradle Enterprise API. Prior to this change, if a build scan was not yet available, the scripts would fail and the experiment summary would not be shown.

### DIFF
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,1 @@
-- [NEW] TBD
+- [FIX] API requests do not allow time for build scans to become available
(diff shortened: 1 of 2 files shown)


release/changes.md
```diff
@@ -1 +1 @@
-- [NEW] TBD
\ No newline at end of file
+- [FIX] API requests do allow time for build scans to become available
```
jthurne, line 1:
```suggestion
- [FIX] Wait up to 10 seconds for Build Scan data to become available
```

jthurne (approved):

release/changes.md
jthurne:
```suggestion
- [FIX] The Build Validation Scripts do not wait long enough for Build Scan data to become available.
```

erichaagdev: I actually missed a word, it should read:

```
- [FIX] API requests do not allow time for build scans to become available
```

This is how we typically write fixes in the release notes, by describing the problem rather than the solution.

etiennestuder (approved):

PR #464, opened 2023-07-11
Added the combine dependabot prs GHA workflow.

ribafish:
Added the combine dependabot prs GHA workflow. As a template I took the workflow file from https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/187

### DIFF
--- a/.github/workflows/combine-dependabot-prs.yml
+++ b/.github/workflows/combine-dependabot-prs.yml
@@ -1,0 +1,23 @@
+name: Combine Dependabot PRs
+
+on:
+  schedule:
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+
+# The minimum permissions required to run this Action
+permissions:
+  contents: write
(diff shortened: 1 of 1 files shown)


runningcode (approved):
Thanks. This looks good.
Just an FYI:
Dependabot themselves just added this feature and it is in beta: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

I say we keep using this method for now.

ribafish:
> Thanks. This looks good. Just an FYI: Dependabot themselves just added this feature and it is in beta: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups
> 
> I say we keep using this method for now.

Yes, I'm aware of that, but since it is in beta I also thought it better to use our workflow until the dependabot feature is stable.

Issue #472, opened 2023-07-19
Dependabot should only update patch version of GE Maven extension in BVS repository


runningcode:

jprinet:
See related PR
https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/476

PR #476, opened 2023-07-24
Update only patch version of GE Maven extension

jprinet:
### Issue
fixes https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/472

### Fix
Use an [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) rule to have Dependabot ignore Major and Minor updates of the GE Maven extension

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,10 @@ updates:
     directory: "/"
     registries:
       - gradle-plugin-portal
+    ignore:
+      # Update only patch version of GE Maven extension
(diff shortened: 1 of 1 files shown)


jprinet:
Is this required for the Maven extension only @runningcode ?

runningcode:
Yes, good to check. We only need to change the maven extension since the other pieces are updated manually.

runningcode (approved):

PR #514, opened 2023-10-05
Updating ccud gradle 

cdsap:
Update:
```
args+=("-Dcom.gradle.enterprise.build-validation.ccud.plugin.version=1.11.3")
```

### DIFF
--- a/components/scripts/lib/gradle.sh
+++ b/components/scripts/lib/gradle.sh
@@ -16,7 +16,7 @@ invoke_gradle() {
 
   if [ "$enable_ge" == "on" ]; then
     args+=("-Dcom.gradle.enterprise.build-validation.gradle.plugin-repository.url=https://plugins.gradle.org/m2")
-    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.15.1")
+    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.14.1")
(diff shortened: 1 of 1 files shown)


components/scripts/lib/gradle.sh
```diff
@@ -16,8 +16,8 @@ invoke_gradle() {
 
   if [ "$enable_ge" == "on" ]; then
     args+=("-Dcom.gradle.enterprise.build-validation.gradle.plugin-repository.url=https://plugins.gradle.org/m2")
-    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.14.1")
-    args+=("-Dcom.gradle.enterprise.build-validation.ccud.plugin.version=1.11.2")
+    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.15")
```
runningcode, line 19: Can you remove this line? The version upgrade process only specifies to update patch versions.

cdsap: just in time, new patch version was released

runningcode: Could you please remove this line since this is a minor version upgrade and the version upgrade process only specifies to update patch versions?
Updating the minor version requires that customers/prospects are using the latest GE server version so we don't want to apply that change.

cdsap: done

PR #523, opened 2023-10-18
Support Develocity API 2023.3.2+ for building experiment summary

erichaagdev:
This PR adds support for Develocity instances running 2023.3.2 and greater when creating the experiment summary. 

This resolves the `ERROR: The required field 'isDisabledDueToError' is not found in the JSON string: {"isEnabled":false}` that is seen when running the scripts with a Develocity instance running 2023.3.2.

Before: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/6543580559/job/17768404455#step:7:84
After: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/6566328643/job/17837095215#step:8:84

### DIFF
--- a/components/fetch-build-scan-data-cmdline-tool/src/main/openapi/openapi.yaml
+++ b/components/fetch-build-scan-data-cmdline-tool/src/main/openapi/openapi.yaml
@@ -1487,7 +1487,6 @@ components:
       description: Information about the local build cache used in the build.
       required:
         - isEnabled
-        - isDisabledDueToError
       properties:
         isEnabled:
           type: boolean
@@ -1503,6 +1502,7 @@ components:
           description: >-
             Whether the local cache was disabled due to an error occuring in
             loading or storing local build cache entries.
+          x-nullable: true
         directory:
           type: string
           description: >-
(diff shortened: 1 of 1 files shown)


release/changes.md
```diff
@@ -1 +1 @@
-- [NEW] TBD
\ No newline at end of file
+[NEW] - Drop support for fetching build scan data from Gradle Enterprise versions older than 2023.3.2
```
erichaagdev, line 1: This is the same wording as when we released `2.0`: https://github.com/gradle/gradle-enterprise-build-validation-scripts/releases/tag/v2.0

etiennestuder:
Shouldn't the field still be in the new version but as optional instead of required?

erichaagdev:
I am not removing the field from the spec if that's what you mean. I have only _removed_ it from the **required:** block.

erichaagdev:
The way GitHub collapses the diff makes it difficult to see, but `isDisabledDueToError` is still there.

image

jthurne (approved):

erichaagdev:
Thanks to the help of @jthurne I was able to test this on a local installation running `2023.3.1` and it worked. ✅

etiennestuder (approved):

PR #524, opened 2023-10-20
Predictive Test Selection is disabled when running builds

### LINKED ISSUE #522: PTS should be disabled when running experiments
When PTS is enabled, test tasks/goals will only store their results in the build cache if build caching is enabled, all tests are selected for execution, and the task/goal was successful. This can throw off the experiment results when PTS decides to skip tests. For this reason we should disable PTS when running the build caching experiments as the focus is on optimizing build caching only.

We can use the system property `pts.enabled` to disable PTS.

erichaagdev:
This PR adds the system property `-Dpts.enabled=false` to build invocations in order to disable PTS. The displayed invocation commands have also been updated to include the system property.  

image

ℹ️ **Note:** The CI/Local experiments (`exp5-gradle` & `exp4-maven`) will also be invoked with `-Dpts.enabled=false` even if the original CI build did not have PTS disabled. 

Closes #522


erichaagdev:
@etiennestuder I want to make sure you see the note I have in the description:

> The CI/Local experiments (exp5-gradle & exp4-maven) will also be invoked with -Dpts.enabled=false even if the original CI build did not have PTS disabled.

I think that is the desired state but it means we need to recommend users also disable PTS in the CI builds. Should we also update the wording then of the interactive mode to include this?

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/5391c5888022de3dae3cf17dec18ce04d69671fc/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh#L276-L287

jthurne (approved):

PR #544, opened 2023-12-15
Updated CCUD gradle plugin to 1.12.1

ribafish:
Updated CCUD gradle plugin to 1.12.1

### DIFF
--- a/components/scripts/lib/gradle.sh
+++ b/components/scripts/lib/gradle.sh
@@ -16,8 +16,8 @@ invoke_gradle() {
 
   if [ "$enable_ge" == "on" ]; then
     args+=("-Dcom.gradle.enterprise.build-validation.gradle.plugin-repository.url=https://plugins.gradle.org/m2")
-    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.14.1")
-    args+=("-Dcom.gradle.enterprise.build-validation.ccud.plugin.version=1.12")
+    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.16.1")
+    args+=("-Dcom.gradle.enterprise.build-validation.ccud.plugin.version=1.12.1")
   fi
 
   if [ -n "${ge_server}" ]; then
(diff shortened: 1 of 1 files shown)


runningcode (changes requested):
I can see how this is confusing in our version upgrade process since this is technically a patch version but this isn't a patch version over the existing version in the codebase. Can you please remove the Develocity Gradle Plugin upgrade from this PR?

components/scripts/lib/gradle.sh
```diff
@@ -16,8 +16,8 @@ invoke_gradle() {
 
   if [ "$enable_ge" == "on" ]; then
     args+=("-Dcom.gradle.enterprise.build-validation.gradle.plugin-repository.url=https://plugins.gradle.org/m2")
-    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.14.1")
-    args+=("-Dcom.gradle.enterprise.build-validation.ccud.plugin.version=1.12")
+    args+=("-Dcom.gradle.enterprise.build-validation.gradle-enterprise.plugin.version=3.14.1") # Max version for GE server 2023.2 -> See https://docs.gradle.com/enterprise/compatibility/#develocity_compatibility
```
runningcode, line 19: This comment is misleading to me since there is no max server version for the GE Gradle plugin 3.14.1. It is compatible with all current and future releases.
I suggest removing the comment along with the link to the compatibility matrix.

ribafish: 3.14.1 is the max supported version for server 2023.2 though.

ribafish: Perhaps I should reword it to `Max version supported by GE server 2023.2 -> See https://docs.gradle.com/enterprise/compatibility/#develocity_compatibility`

erichaagdev:
@runningcode @ribafish I think we can close this in favor of #573.

ribafish:
> @runningcode @ribafish I think we can close this in favor of #573.

@erichaagdev yeah, sure. @runningcode I still think that we should add a comment about ge-plugin version, waht do you think?

erichaagdev:
I'm going to close this. The comment about the Develocity plugin version is only relevant to us when there is a new plugin release. If we feel we need to add something about it, it should be added to our version upgrades process and not as a code comment.

PR #579, opened 2024-03-12
Configure both GE and Develocity extensions available in the evaluated project

pshevche:
## Summary
Maven validation scripts interact with the extension using system properties (primarily for cache configuration). In this PR, we now set properties both for the GE and Develocity extension, making it forwards compatible with extension versions available under the Develocity name.


pshevche:
This misses a couple pieces

PR #612, opened 2024-05-03
Upgrade Shellcheck to v0.10.0

jthurne:
Dependabot does not upgrade Shellcheck because it isn't declared like a normal dependency.

Also addresses [SC2004](https://www.shellcheck.net/wiki/SC2004), which was detected by the new version of shellcheck.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,7 +55,7 @@ dependencies {
 
 shellcheck {
     additionalArguments = "-a -x"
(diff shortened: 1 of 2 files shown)


erichaagdev (approved):
I ran the build locally and it was successful. 👍 

https://ge.solutions-team.gradle.com/s/c5t5ltxi3np4y/timeline?type=com.felipefzdz.gradle.shellcheck.Shellcheck

PR #624, opened 2024-05-28
Add support for the renamed Develocity Maven Extension

jthurne:
The Gradle Enterprise Maven Extension was renamed to the Develocity Maven
Extension in version 1.21.  The API was also updated for hte product name
change. Unfortunately, this broke compatibility with the Build Validation
Scripts. This commit fixes the incompatibility.

The fix works by updating the configure-gradle-enterprise-maven-extension to
define and register two listeners: a GradleEnterpriseListener and a
DevelocityListener.  If an older version of the Gradle Enterprise Maven
Extension is used, it will load the GradleEnterpriseListener.  If a newer
version of the Develocity Maven Extension is used, then it will load the
DevelocityListener. The Develocity Maven Extension Adapters are used to
eliminate duplication between the two different listeners.

To ensure compatibility with older verisons of Maven, this PR also
updates the extension to be compiled against an earlier version of
maven-core.

### DIFF
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("java")
+    id("com.github.johnrengelman.shadow") version "8.1.1"
 }
 
 repositories {
@@ -10,7 +11,7 @@ dependencies {
     compileOnly("org.apache.maven:maven-core:3.9.6")
     compileOnly("org.codehaus.plexus:plexus-component-annotations:2.2.0")
     compileOnly("com.gradle:develocity-maven-extension:1.21.3")
-    implementation("com.gradle:develocity-maven-extension-adapters:${rootProject.extra["develocityMavenExtensionAdaptersVersion"]}")
+    implementation("com.gradle:develocity-maven-extension-adapters:1.0")
 }
 
 description = "Maven extension to capture the build scan URL"
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -25,4 +25,5 @@ java {
 
 tasks.withType(JavaCompile::class).configureEach {
     options.encoding = "UTF-8"
+    options.compilerArgs.add("-Xlint:deprecation")
 }
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -9,7 +9,8 @@ repositories {
 dependencies {
     compileOnly("org.apache.maven:maven-core:3.9.6")
     compileOnly("org.codehaus.plexus:plexus-component-annotations:2.2.0")
-    compileOnly("com.gradle:gradle-enterprise-maven-extension:1.18.4")
+    compileOnly("com.gradle:develocity-maven-extension:1.21.3")
+    implementation("com.gradle:develocity-maven-extension-adapters:${rootProject.extra["develocityMavenExtensionAdaptersVersion"]}")
 }
 
 description = "Maven extension to capture the build scan URL"
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -10,7 +10,7 @@ repositories {
 dependencies {
     compileOnly("org.apache.maven:maven-core:3.6.3")  // intentionally compiling against an older version to preserve compatibility with older versions of Maven
     compileOnly("org.codehaus.plexus:plexus-component-annotations:2.2.0")
-    compileOnly("com.gradle:develocity-maven-extension:1.21.3")
+    compileOnly("com.gradle:develocity-maven-extension:1.21.4")
     implementation("com.gradle:develocity-maven-extension-adapters:1.0")
 }
 
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -25,5 +25,4 @@ java {
 
 tasks.withType(JavaCompile::class).configureEach {
     options.encoding = "UTF-8"
-    options.compilerArgs.add("-Xlint:deprecation")
 }
--- a/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
+++ b/components/configure-gradle-enterprise-maven-extension/build.gradle.kts
@@ -8,7 +8,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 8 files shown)


components/configure-gradle-enterprise-maven-extension/build.gradle.kts
```diff
@@ -1,15 +1,17 @@
...
 repositories {
     mavenCentral()
 }
 
 dependencies {
-    compileOnly("org.apache.maven:maven-core:3.9.7")
+    compileOnly("org.apache.maven:maven-core:3.6.3")  // intentionally compiling against an older version to preserve compatibility with older versions of Maven
     compileOnly("org.codehaus.plexus:plexus-component-annotations:2.2.0")
-    compileOnly("com.gradle:gradle-enterprise-maven-extension:1.18.4")
+    compileOnly("com.gradle:develocity-maven-extension:1.21.3")
```
runningcode, line 13: Should this be `testCompileOnly`or removed to prevent accidental direct usages of this instead of the adapters?

runningcode (commented):
just a small question otherwise looks good.

components/configure-gradle-enterprise-maven-extension/build.gradle.kts
erichaagdev: Nice catch. There are no tests so it should be removed.

jthurne: Unfortunately, if this is removed, the project doesn't compile: https://ge.solutions-team.gradle.com/s/etr5srswtsk6y

It appears the develocity-maven-extension-adapters doesn't declare the develocity-maven-extension as a dependency.

We need classes in that library to compile because the ConfigureDevelocity and ConfigureGradleEnterprise listeners implement interfaces from that library.

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureGradleEnterprise.java
erichaagdev, line 1: Same suggestions to this file as were made for: components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureDevelocityAdaptor.java

erichaagdev (changes requested):

components/configure-gradle-enterprise-maven-extension/build.gradle.kts
erichaagdev: I guess that makes sense to avoid classpath conflicts in the adapters?

cc @pshevche

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureDevelocity.java
```diff
@@ -0,0 +1,27 @@
...
+
+@SuppressWarnings("unused")
+public class ConfigureDevelocity implements DevelocityListener {
+
+    private static final String EXPERIMENT_DIR = System.getProperty("com.gradle.enterprise.build-validation.expDir");
+
+    private final ConfigureDevelocityAdaptor configureDevelocityAdaptor;
+
+    @Inject
+    public ConfigureDevelocity(ConfigureDevelocityAdaptor configureDevelocityAdaptor, RootProjectExtractor rootProjectExtractor, Logger logger) {
```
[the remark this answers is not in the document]
jthurne: Done in https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/624/commits/f363d45a29e0cf1975acb3270e6575c9095fe8bb

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureDevelocity.java
```diff
@@ -0,0 +1,27 @@
...
+import com.gradle.develocity.agent.maven.api.DevelocityListener;
+import org.apache.maven.execution.MavenSession;
+import org.codehaus.plexus.logging.Logger;
+
+import javax.inject.Inject;
+
+@SuppressWarnings("unused")
+public class ConfigureDevelocity implements DevelocityListener {
+
+    private static final String EXPERIMENT_DIR = System.getProperty("com.gradle.enterprise.build-validation.expDir");
```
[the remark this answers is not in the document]
jthurne: Done in https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/624/commits/f363d45a29e0cf1975acb3270e6575c9095fe8bb

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureDevelocity.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.gradle.develocity.agent.maven.adapters.develocity.DevelocityApiAdapter;
+import com.gradle.develocity.agent.maven.api.DevelocityApi;
+import com.gradle.develocity.agent.maven.api.DevelocityListener;
+import org.apache.maven.execution.MavenSession;
+import org.codehaus.plexus.logging.Logger;
+
+import javax.inject.Inject;
+
+@SuppressWarnings("unused")
```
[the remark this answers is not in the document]
jthurne: It turns out I'm wrong. The compiler is not configured to fail on warnings, and unused warnings only show up in the IDE anyway.  I removed it in https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/624/commits/f363d45a29e0cf1975acb3270e6575c9095fe8bb

components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureGradleEnterprise.java
jthurne: Done in https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/624/commits/f363d45a29e0cf1975acb3270e6575c9095fe8bb

components/configure-gradle-enterprise-maven-extension/build.gradle.kts
jthurne: > I guess that makes sense to avoid classpath conflicts in the adapters?

That would make sense. It's the same reason why we declare it as `compileOnly` instead of `implementation`: to avoid version conflicts with whatever version of the extension the user is actually using.

runningcode: ah that makes sense. good to keep it then.

erichaagdev (approved):
I did some testing and everything checks out. Here are the scenarios I ran:

image

The column in red shows what the build has applied already. 

I even went so far as to update the scripts to inject DV 1.21.4, based off of this branch, and everything worked as expected.

components/configure-gradle-enterprise-maven-extension/build.gradle.kts
bdemers:
```suggestion
    compileOnly("com.gradle:develocity-maven-extension:1.21.4")
```

PR #629, opened 2024-06-13
Move Maven jars to avoid any licensing confusion

jthurne:
- **Move the Develocity Maven extension to lib/develocity**.  It is covered by the Develocity terms of service.
- **Move common-custom-user-data-maven-extension to lib/third-party**.  From the perspective of this project, it is another third-party OSS dependency.


components/scripts/lib/maven.sh
```diff
@@ -1,6 +1,17 @@
 #!/usr/bin/env bash
 
-readonly CONFIGURE_GRADLE_ENTERPRISE_JAR="${LIB_DIR}/maven-libs/configure-gradle-enterprise-maven-extension-${SCRIPT_VERSION}.jar"
+find_versioned_jar() {
+  lcoal dir_to_search base_name
```
erichaagdev, line 4:
```suggestion
  local dir_to_search base_name
```

erichaagdev (changes requested):

erichaagdev (approved):

PR #633, opened 2024-06-20
Github actions access token migration

ribafish:
- Migrated GitHub actions gradle executions to access token auth 
- Migrated GitHub actions cross build tests of BVS env var access key to develocity naming

### DIFF
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -19,7 +19,7 @@ jobs:
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v3
+        with:
+          develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
       - name: Build with Gradle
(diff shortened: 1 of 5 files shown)


erichaagdev (approved):

.github/workflows/build-verification.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v3
+        with:
+          develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
       - name: Build with Gradle
         run: ./gradlew build -i
```
jprinet, line 25: This is not a change you introduced, but I'm wondering if logging at `info` level is necessary

jprinet (approved):
LGTM

.github/workflows/build-verification.yml
erichaagdev: I think it can be removed.

Issue #644, opened 2024-08-02
Artifact name for 'Archive receipt' should be configurable


wbonnefond:
### Issue:

When running the experiments using a [matrixed workflow](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs), the `Archvie receipt` step will fail because the artifact name is the same across all matrixed jobs in the workflow. This was a breaking change with the [v4 upload artifact release.](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes)

<img width="1298" alt="Screenshot 2024-08-02 at 10 57 04 AM" src="https://github.com/user-attachments/assets/cd40b5cf-546e-4d50-be2e-bb66548102d0">

### Solution

The artifact name for the `Archive receipt` step for each experiment should be configurable as part of the action

erichaagdev:
Hi, @wbonnefond. Thanks for taking the time to report this. 

Do you actually need full control over the artifact name or would it be acceptable if we could ensure the artifact name was unique across all matrixed runs?

PR #645, opened 2024-08-02
Fix for #644, allow configuration of artifact name for `Archive receipt` step

wbonnefond:
To ensure the `Archive receipt` step works when run in a matrixed workflow, the artifact name needs to be unique.

This change adds an input to the action for the artifact name `archieveReceiptArtifactName` and defaults to the previous hardcoded value. This allows callers to specify the name if needed. 

In our usecase, we run this in a matrixed workflow against 8 apps, so we can ensure the name is different by appending the app name to the artifact name like:

```
- name: Run experiment 1
  uses: gradle/gradle-enterprise-build-validation-scripts/.github/actions/gradle/experiment-1@actions-stable
  env:
    DEVELOCITY_ACCESS_KEY: xxxxxx
  with:
    gitRepo: ${{ env.GIT_REPO }}
    gradleEnterpriseUrl: ${{ env.GRADLE_ENTERPRISE_URL }}
    tasks: ${{ env.TASKS }}
    projectDir: ${{ matrix.target.project }}
    archiveReceiptArtifactName: "experiment-1-receipt-${{ matrix.target.project }}"
```

### DIFF
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -29,6 +29,10 @@ inputs:
   enableGradleEnterprise:
     description: "Enables Gradle Enterprise on a project not already connected"
     required: false
+  archiveRecieptArtifactName:
+    description: "The artifact name for the archive step"
+    default: "experiment-1-receipt"
+    required: false
 outputs:
   buildScanFirstBuild:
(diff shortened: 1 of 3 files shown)


wbonnefond:
Closing as this is being fixed in #648

PR #648, opened 2024-08-05
Use unique artifact identifier in upload-artifact step

jprinet:
### Issue
Fixes https://github.com/gradle/gradle-enterprise-build-validation-scripts/issues/644

Since [upload-artifact@v4](https://github.com/actions/upload-artifact), uploading again the same artifact name in a workflow fails the step.
Using a hardcoded `experiment-X-receipt` is not valid anymore

### Fix
Use a unique name. This is not so straightforward as some collisions can occur in the case of matrix.
Using the `job-index` for matrix solves this.
Note that appending the `job-index` is happening only if `job-total > 1` to not suffix artifacts from non matrix scenario

This could hopefully get simplified if/when GitHub starts to expose a unique `job-id` when running with matrix strategy.

### Extra
A receipt download link has been added as part of the workflow summary

### Example
See a summary running with the PR fix [here](https://github.com/jprinet/matrix-workflow-test/actions/runs/10268393419)

### DIFF
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -110,5 +110,5 @@ runs:
     - name: Archive receipt
       uses: actions/upload-artifact@v4
       with:
-        name: experiment-1-receipt-${{ github.job }}${{ strategy.job-index && format('-{}', strategy.job-index) || '' }}
+        name: experiment-1-receipt-${{ github.job }}-${{ strategy.job-index || '0' }}
         path: gradle-enterprise-gradle-build-validation/.data/01-validate-incremental-building/latest*/exp1-*.receipt
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -110,5 +110,5 @@ runs:
     - name: Archive receipt
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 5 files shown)


wbonnefond:
This branch fixes the issue I was having in our codebase

.github/actions/gradle/experiment-1/action.yml
```diff
@@ -108,7 +108,13 @@ runs:
...
+      id: upload-artifact
       uses: actions/upload-artifact@v4
       with:
-        name: experiment-1-receipt
+        name: experiment-1-receipt-${{ github.job }}${{ strategy.job-total > 1 && format('-{0}', strategy.job-index) || '' }}
         path: gradle-enterprise-gradle-build-validation/.data/01-validate-incremental-building/latest*/exp1-*.receipt
+    - name: Add artifact link to summary
+      run: |
+        echo "-------------" >> $GITHUB_STEP_SUMMARY
+        echo "Download receipt: [${{ steps.upload-artifact.outputs.artifact-url }}](${{ steps.upload-artifact.outputs.artifact-url }})" >> $GITHUB_STEP_SUMMARY
```
erichaagdev, line 119: It's already a link. Is it necessary to wrap it like this?

```suggestion
        echo "Download receipt: ${{ steps.upload-artifact.outputs.artifact-url }}" >> $GITHUB_STEP_SUMMARY
```

jprinet: Good point 😅
I adjusted the PR

erichaagdev (approved):
🚀

PR #663, opened 2024-09-13
Instructions include callout about committing for local Git repositories

erichaagdev:
This adds a quick note to the READMEs that changes must still be committed when using a local Git repository.

### DIFF
--- a/Gradle.md
+++ b/Gradle.md
@@ -247,6 +247,10 @@ git clone https://github.com/gradle/gradle-build-scan-quickstart $HOME/gradle-bu
 ./02-validate-local-build-caching-same-location.sh -i -r file://$HOME/gradle-build-scan-quickstart
 ```
 
+> [!IMPORTANT]
+> Regardless of whether you use a local or remote Git repository, any changes _must be_ commited for them to be picked up by the experiments.
+> The benefit of using a local Git repository is that the changes don't need to be pushed to the remote repository.
+
 ## Dealing with filename too long errors on Windows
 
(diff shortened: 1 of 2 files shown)


Gradle.md
```diff
@@ -247,6 +247,10 @@ git clone https://github.com/gradle/gradle-build-scan-quickstart $HOME/gradle-bu
 ./02-validate-local-build-caching-same-location.sh -i -r file://$HOME/gradle-build-scan-quickstart
 ```
 
+> [!IMPORTANT]
+> Regardless of whether you use a local or remote Git repository, any changes _must be_ commited for them to be included in the experiment.
```
etiennestuder, line 251: or `...for them to be picked up by the experiments.` ?

Gradle.md
```diff
@@ -247,6 +247,10 @@ git clone https://github.com/gradle/gradle-build-scan-quickstart $HOME/gradle-bu
...
+> The benefit of using a local Git repository is that the changes don't also need to be pushed to the remote.
```
etiennestuder, line 252: or `...that the changes don't need to be pushed to the remote repository.`

Maven.md
```diff
@@ -245,6 +245,10 @@ git clone https://github.com/gradle/maven-build-scan-quickstart $HOME/maven-buil
 ./01-validate-local-build-caching-same-location.sh -i -r file://$HOME/maven-build-scan-quickstart
 ```
 
+> [!IMPORTANT]
+> Regardless of whether you use a local or remote Git repository, any changes _must be_ commited for them to be included in the experiment.
+> The benefit of using a local Git repository is that the changes don't also need to be pushed to the remote.
+
```
etiennestuder, line 251: Same comments as above for Gradle.

etiennestuder (commented):
Thanks. I added some proposal to maybe make it a tiny bit more easy to consume.

PR #671, opened 2024-10-21
Common Develocity injection script is used for Gradle experiments

erichaagdev:
This PR changes the BVS to leverage the tried and true common [Develocity injection script](https://github.com/gradle/develocity-ci-injection) to configure Develocity.

The approach taken was to use the common injection script where possible and move any BVS-specific logic to a secondary init script which I've called `configure-build-validation.gradle`. Both the common injection script and the BVS-specific script are applied to all Gradle invocations and are configured via system properties. 

By leveraging the common injection script, we can ensure we don't break injection functionality, which in the past has proven to be somewhat fragile.


erichaagdev:
Successful cross platform test run: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/11524907372

etiennestuder (approved):
Great work.

PR #673, opened 2024-10-25
All user facing components are renamed to Develocity

erichaagdev:
This PR may seem daunting but I've broken it up into several very focused commits that should be easy to follow. I suggest stepping through one commit at a time.

1. 1cc79b2d05a3f33c8d2cd98c648eaedd16f55970 updates all user facing text, e.g., interactive prompts, README. No functional changes were made and it should be easy to verify
2. 7604767177781567344a30883ccf7757f11919f1 adds Develocity parameters to all GitHub Actions workflows and deprecates Gradle Enterprise parameters
3. 87102d0d8a1a036d9e5e4c95d9557ddd1798a2e9 adds Develocity script arguments and deprecates Gradle Enterprise arguments
4. 9ab1005527f0cf9c992527d6246211f31438f1ea adds Develocity branded distributions and deprecates Gradle Enterprise distributions

> [!IMPORTANT]
> All changes are backwards compatible. Aside from user facing text, nothing relating to Gradle Enterprise was removed.

When using the legacy `--enable-gradle-enterprise`, `--gradle-enterprise-server`, or `gradle-enterprise-x-build-validation` distribution, a deprecation warning will be displayed in the experiment summary.

image

`--enable-gradle-enterprise` and `--gradle-enterprise-server` are no longer displayed in the `--help` text, although it's still possible to use them. `-s` is now an alias for `--develocity-server` and `-e` is now an alias for `--enable-develocity`.

image

In the case where both a Gradle Enterprise and Develocity parameter are used, the Develocity parameter will take precedence. This is true for both the GitHub Actions workflows and the command line arguments. 

The build now produces 4 distributions. For each build tool there is a Develocity and Gradle Enterprise distribution. A GitHub release will release all 4 artifacts. 

image


etiennestuder (approved):
Great work, too.

PR #675, opened 2024-11-05
Use environment variables to configure Gradle init scripts

erichaagdev:
System properties cannot be read reliably from init scripts for older versions of Gradle. This is not the case for environment variables. This PR changes the configuration of the init scripts to be done using environment variables instead of system properties.

### DIFF
--- a/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
+++ b/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
@@ -194,19 +194,18 @@ validate_build_config() {
 }
 
 execute_build() {
-  local args
-  args=(--build-cache --init-script "${INIT_SCRIPTS_DIR}/configure-remote-build-caching.gradle")
-  if [ -n "${remote_build_cache_url}" ]; then
-    args+=("-Ddevelocity.build-validation.remoteBuildCacheUrl=${remote_build_cache_url}")
-  fi
-
-  # shellcheck disable=SC2206  # we want tasks to expand with word splitting in this case
-  args+=(clean ${tasks})
-
   info "Running build:"
-  info "./gradlew --build-cache -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} -Dpts.enabled=false clean ${tasks}$(print_extra_args)"
+  print_gradle_command
 
-  invoke_gradle 1 "${args[@]}"
+  # shellcheck disable=SC2086  # we want tasks to expand with word splitting in this case
+  invoke_gradle 1 \
+     --build-cache \
+     --init-script "${INIT_SCRIPTS_DIR}/configure-remote-build-caching.gradle" \
+     clean ${tasks}
+}
+
+print_gradle_command() {
+  info "./gradlew --build-cache -Dscan.tag.${EXP_SCAN_TAG} -Dscan.value.runId=${RUN_ID} -Dpts.enabled=false clean ${tasks}$(print_extra_args)"
 }
 
 # Overrides summary.sh#print_experiment_specific_summary_info
--- a/components/scripts/lib/gradle.sh
+++ b/components/scripts/lib/gradle.sh
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 
 invoke_gradle() {
-  local run_num args
-  args=()
+  local run_num
   run_num=$1
   shift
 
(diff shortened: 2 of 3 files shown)


erichaagdev:
Successful cross platform test run: https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/11686512948/job/32542510765

components/scripts/lib/gradle.sh
```diff
@@ -57,9 +66,13 @@ invoke_gradle() {
...
   debug "Current directory: $(pwd)"
-  debug ./gradlew "${args[@]}"
+  # shellcheck disable=SC2145
+  debug export "${envs[@]}"';' ./gradlew "${args[@]}"
 
-  if ./gradlew "${args[@]}"; then
+  # The parenthesis below will intentionally create a subshell. This causes the
+  # environment variables to only be exported for the child process and not leak
+  # to the rest of the script.
+  if (export "${envs[@]}"; ./gradlew "${args[@]}"); then
```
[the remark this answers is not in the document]
erichaagdev: The scripts only support Bash environments so there's no need to worry about Powershell. The cross platform tests verify it works on Windows with WSL. https://github.com/gradle/gradle-enterprise-build-validation-scripts/actions/runs/11686512948/job/32542519317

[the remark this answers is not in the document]
jprinet: Perfect then 👍

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -194,19 +194,18 @@ validate_build_config() {
 }
 
 execute_build() {
-  local args
-  args=(--build-cache --init-script "${INIT_SCRIPTS_DIR}/configure-remote-build-caching.gradle")
-  if [ -n "${remote_build_cache_url}" ]; then
-    args+=("-Ddevelocity.build-validation.remoteBuildCacheUrl=${remote_build_cache_url}")
-  fi
```
jthurne, line 201: Was this replaced by an env var somewhere? If yes, I don't immediately see where.

components/scripts/lib/gradle.sh
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 
 invoke_gradle() {
-  local run_num args
-  args=()
+  local run_num
```
jthurne, line 4: I suggest declaring envs as a local variable as well.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
erichaagdev: Yes. I had to move it to `gradle.sh` because I didn't see a way to make `invoke_gradle` accept both extra envs and extra args.

erichaagdev: https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/71fe28813cad8756e73f88365190947d17baf3b9/components/scripts/lib/gradle.sh#L34-L38

components/scripts/lib/gradle.sh
erichaagdev: Good idea. Done in 95707cff830c11bb607d5a18abc844c87349330d.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
erichaagdev: Only experiment 5 exposes an option to configure the remote build cache URL, but I can't think of a reason it would cause problems moving it to a common function.

jthurne: Ok, that makes sense.  👍

jthurne (approved):

components/scripts/lib/gradle.sh
```diff
@@ -12,35 +12,45 @@ invoke_gradle() {
...
+      DEVELOCITY_BUILD_VALIDATION_URL="${ge_server}"
+      DEVELOCITY_BUILD_VALIDATION_ALLOW_UNTRUSTED_SERVER=false
+    )
+  fi
+
+  if [ -n "${remote_build_cache_url}" ]; then
+    envs+=(
+      DEVELOCITY_BUILD_VALIDATION_REMOTEBUILDCACHEURL="${remote_build_cache_url}"
     )
   fi
```
etiennestuder, line 39: This `if` block is new. Can you explain?

etiennestuder: I now understand.

etiennestuder (approved):
LGTM.

etiennestuder:
@erichaagdev Regarding _System properties cannot be read reliably from init scripts for older versions of Gradle._:

I don’t agree with that. It’s only since changing to `System.properties` (in the shared init script) that we are facing issues. It worked well in all scenarios when we were using `startParameters...`.

components/scripts/lib/gradle.sh
erichaagdev: 👍 Jim brought it up too and it was discussed here, in case you missed it. 

https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/675#discussion_r1829896368

PR #676, opened 2024-11-05
Expose option to download latest development release in the download actions

erichaagdev:
This PR adds an additional optional parameter to download the latest development release of the scripts. If left unspecified, the latest released version will be downloaded.

### DIFF
--- a/.github/actions/gradle/download/action.yml
+++ b/.github/actions/gradle/download/action.yml
@@ -5,6 +5,10 @@ inputs:
   token:
     description: "GitHub token"
     required: false
+  downloadDevelopmentRelease:
+    description: "Downloads the latest development version when 'true'; by default, the latest released version is downloaded"
+    default: "false"
+    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


.github/actions/gradle/download/action.yml
```diff
@@ -19,8 +22,13 @@ runs:
           curlOpts+=(--header 'Authorization: Bearer ${{ inputs.token }}')
         fi
 
+        releaseUrl="https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest"
+        if [ ! -z "${{ inputs.version }}" ]; then
+          releaseUrl="https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/tags/${{ inputs.version }}"
+        fi
+
         # Get the download url of the latest release
```
etiennestuder, line 30: The comment should be adjusted (it's not always getting the latest release)

.github/actions/maven/download/action.yml
```diff
@@ -19,8 +22,13 @@ runs:
           curlOpts+=(--header 'Authorization: Bearer ${{ inputs.token }}')
         fi
 
+        releaseUrl="https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/latest"
+        if [ ! -z "${{ inputs.version }}" ]; then
+          releaseUrl="https://api.github.com/repos/gradle/gradle-enterprise-build-validation-scripts/releases/tags/${{ inputs.version }}"
+        fi
+
         # Get the download url of the latest release
```
etiennestuder, line 30: The comment should be adjusted (it's not always getting the latest release)

etiennestuder (changes requested):

jprinet (changes requested):
With this change, we could run some gradle-enterprise based version or some develocity based version, which conditions the folder name containing the script.
This has to be addressed for each build tool and each experiments.
See as an example for Gradle on experiment 1 [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/5a22aff39b59b2bc9f2e2422243047f1b234a375/.github/actions/gradle/experiment-1/action.yml#L99) and [here](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/5a22aff39b59b2bc9f2e2422243047f1b234a375/.github/actions/gradle/experiment-1/action.yml#L129)

erichaagdev:
@jprinet that is a very good point. I see two options:

1) Make this work for both DV and GE distributions, which will complicate logic a bit
2) Change this approach to only expose an option to download the latest development version

I am leaning towards #2 until we have a use case for #1. So, we would have a boolean parameter like `useDevelopmentVersion` that's false by default and you would not have a way to control the version. 

What do you think?

jprinet:
> 1. Make this work for both DV and GE distributions, which will complicate logic a bit
> 2. Change this approach to only expose an option to download the latest development version
> 
> I am leaning towards # 2 until we have a use case for # 1. So, we would have a boolean parameter like `useDevelopmentVersion` that's false by default and you would not have a way to control the version.
> 
> What do you think?

I agree with your proposal, this is what we want at the moment and it won't pollute all the composite actions with repeated tedious logic

.github/actions/gradle/download/action.yml
erichaagdev: Good point. I've updated this in the latest commit. 831909bfd1b9ce6fc7306ad65e51953f06519b3a

erichaagdev:
@jprinet I've updated the PR according to our discussion.

jprinet (approved):
LGTM 👍
There is a [README page](https://github.com/gradle/gradle-enterprise-build-validation-scripts/tree/main/.github/actions/gradle) for each build tool, but I think it's fair to not document this parameter

erichaagdev:
Going to merge this.

PR #680, opened 2024-11-12
Retain pre-existing remote build cache config

### LINKED ISSUE #507: Build Validation Scripts do not configure Gradle Enterprise or Develocity remote cache connector
The Build Validation Scripts do not appear to retain configuration specific to the GE remote cache connector or the GE plugin.

Here are the two specific instances of this issue I've come across:
1. `allowInsecureProtocol` was set to `true`, but the local build for Experiment 5 failed with `Using insecure protocols with remote build cache, without explicit opt-in, is unsupported`. The workaround for this was to add `allowInsecureProtocol=true` to the validation scripts' remote cache init script.
2. `accessKey` was configured through the GE plugin extension, but the local build for Experiment 5 was failing to authenticate with the remote cache. This was resolved by configuring the access key in `keys.properties`

Ideally all GE configuration applied to the project would be carried over to the script executions.

(Note that resolving the `allowInsecureProtocol` issue would require updating the GE API to include the value of `allowInsecureProtocol`.)

gabrielfeo:
Retain pre-existing remote cache configuration when setting the cache path, by never replacing an existing cache. Also support using the Develocity connector and warn in case the build has a custom cache type, for which the script cannot set the cache path.

Closes #507

### DIFF
--- a/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
@@ -11,11 +11,21 @@ settingsEvaluated { settings ->
         local {
             enabled = false
         }
-        remote(HttpBuildCache) {
+        (remote ?: remote(HttpBuildCache)).with {
             enabled = true
             push = false
-            if (remoteBuildCacheUrl) {
-                url = withPathTrailingSlash(new URI(remoteBuildCacheUrl))
+            def uri = remoteBuildCacheUrl
+                ? withPathTrailingSlash(new URI(remoteBuildCacheUrl))
+                : null
+            if (uri && delegate instanceof HttpBuildCache) {
+                url = uri
+            } else if (uri && delegate.class.name.startsWith('com.gradle.develocity')) {
+                path = uri.path
+            } else {
+                logger.warn("""
+                    Couldn't set build cache path automatically for remote of type '${delegate.class.name}'.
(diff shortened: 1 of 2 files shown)


erichaagdev:
Hi, @gabrielfeo. 👋 

Thank you very much for your continued interest in improving the Develocity Build Validation Scripts.

What makes this more challenging is we must also continue to support the Gradle Enterprise plugin. 

Prior to invoking the build, we call the Develocity API to query for several custom values in order to "recreate" the build as it was run in CI. This includes values like the repository, commit ID, branch, etc. You can see the full list here.

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/17e8857bdbee720fe5ff4732f7f667d80fb94a03/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh#L162-L180

How I've always wanted to handle this is by updating the API call to return the configured remote build cache type. This is [visible in a build scan](https://ge.solutions-team.gradle.com/s/odv6bme256wso/performance/build-cache#remote-cache-type) and [available in the Develocity API](https://docs.gradle.com/enterprise/api-manual/ref/2024.2.html#tag/Builds/operation/GetGradleBuildCachePerformance). The remote build cache can be configured accordingly from there. Unfortunately, the API client used by the scripts is closed source so this couldn't be implemented by an external contributor.

That being said, I will go ahead and leave this PR open to be discussed with the team.

PR #682, opened 2024-11-20
Environment variable changes are reverted and `gradle.startParameter` is used to read system properties in init scripts

erichaagdev:
This PR:

1) Reverts the change for configuring init scripts via environment variables
2) (re)Migrates init scripts to use `gradle.startParameter` to read system properties

I suggest reviewing only the last 2 commits as the first 2 only revert the previous changes.

Relevant diff: https://github.com/gradle/gradle-enterprise-build-validation-scripts/pull/682/files/64f48ec26c9fb43c51d62d281a9c3a99554342e1..632519d8c0cf14190fe3ac6608b72ac50fd791d4

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -167,6 +167,7 @@ val copyGradleScripts by tasks.registering(Sync::class) {
     from(develocityInjectionResolvable) {
         rename { "develocity-injection.gradle" }
         into("lib/scripts/gradle-init-scripts")
+        filter(TransformDevelocityInjectionScript())
     }
     from(layout.projectDirectory.dir("components/scripts")) {
         include("README.md")
--- a/buildSrc/src/main/kotlin/TransformDevelocityInjectionScript.kt
+++ b/buildSrc/src/main/kotlin/TransformDevelocityInjectionScript.kt
@@ -1,0 +1,27 @@
+/*
+ * Transforms usages of 'System.getProperty' in the Develocity injection script
+ * to 'gradle.startParameter'. This is required because 'System.getProperty'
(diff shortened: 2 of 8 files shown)


jthurne (approved):

bigdaz (approved):
Since these changes _are_ required for BVS, is there a good reason why the same problem won't manifest for other init-script consumers?

I'm OK with these changes to BVS, but I'd request changes to apply these to the reference init-script. Namely, I think we can avoid the duplicate definition of `getInputParam(name)` by changing the signature to `getInputParam(gradle, name)`.

erichaagdev:
> is there a good reason why the same problem won't manifest for other init-script consumers?

I expect this is a universal issue. In earlier versions of Gradle, what I discovered is that any modification to environment variables (changes, additions, etc.) are not reflected throughout the life of the Daemon process. It will be more common that you have Daemon reuse locally than in CI.

The other consideration is locally these environment variables are only included when invoking the scripts. In CI, the environment variables are always present in CI for every build invocation, though modifications are possible, it will be very infrequent.

I think this essentially makes it a non-issue for CI plugins. I believe this is the relevant Gradle issue: https://github.com/gradle/gradle/issues/10483

> I'd request changes to apply these to the reference init-script. Namely, I think we can avoid the duplicate definition of getInputParam(name) by changing the signature to getInputParam(gradle, name).

I can follow up with a PR. 👍

etiennestuder (commented):
LGTM. @erichaagdev and I went through the PR review together.

etiennestuder (approved):

Issue #777, opened 2024-12-20
Fetching Build Scan data for all builds dumps 404 error pages in logs


SimonMarquis:
When using the default Gradle Scans endpoint, API calls are not allowed.
Leading to 404 errors when the scripts are trying to fetch all the build scan data.
Here is an example: https://github.com/SimonMarquis/SealedObjectInstances/actions/runs/12433888640/job/34716406872?pr=313#step:5:452

Unfortunately, the whole HTML page is then dumped to the logs.
Is this on purpose? If not, can we prevent to dump this useless content to the logs?

Thanks

erichaagdev:
Hi, @SimonMarquis. Thanks for taking the time to file this issue. 

The Develocity Build Validation Scripts are not intended to be used with `https://scans.gradle.com`. We will discuss internally to see if we can make this experience better.

erichaagdev:
Hi @SimonMarquis. As you already pointed out, API calls are not allowed to [scans.gradle.com](https://scans.gradle.com), and we don't have any plans to change this. 

The typical use case for running these in CI is to automatically catch build caching regressions by using the `failIfNotFullyCacheable` option. This option only works with access to the Develocity API which we use to determine task outcomes. Therefore, even if we improved the experience of using the scripts with [scans.gradle.com](https://scans.gradle.com), we will never be able to make `failIfNotFullyCacheable` work with it and you would need to manually check the resulting build scans yourself. 

Could you help me understand your use case for running the Develocity Build Validation Scripts? Is it to catch build caching regressions?

SimonMarquis:
> Could you help me understand your use case for running the Develocity Build Validation Scripts? Is it to catch build caching regressions?

Yes, exactly.
I fully understand the limitations of the public scans endpoint and that it is not meant to be targeted through API calls.
The only issue I had was that I was trying to debug something, and I was presented the entire HTML dump of the webpage in the logs. And this is probably not useful to anyone (a simple error message with the response message and error code would suffice) and makes the debugging experience slightly worse than it needs to be.

If this is not planned or if you find this would need to support too many edge cases, then I'll be fine closing this issue.

erichaagdev:
As we discussed in the PR, here is a separate issue to add support for `scans.gradle.com`.

https://github.com/gradle/develocity-build-validation-scripts/issues/795

PR #788, opened 2025-01-24
Develocity is added as an explicit requirement

erichaagdev:
The purpose of this PR is to clarify that Develocity is required in order to use the scripts.


Gradle.md
```diff
@@ -8,6 +8,8 @@ There are currently five experiments for Gradle. You could also perform these ex
 
 ## Requirements
 
+You must have [Develocity](https://gradle.com/develocity/) in order to use the Develocity Build Validation Scripts. At a minimum, your Develocity user must have the permissions `Publish build scans` and `View build scans and build data` granted explicitly or anonymously, as analyzing the results relies on the use of build scans.
```
clayburn, line 11: This is correct, although I wonder if we should mention the benefit of having API here as well

clayburn (approved):
LGTM - take or leave the comment

Gradle.md
erichaagdev: I thought about that, but ultimately decided not to since it isn't actually a "requirement". We also mention the API [here](https://github.com/gradle/develocity-build-validation-scripts/blob/main/Gradle.md#authenticating-with-develocity).

jthurne:
```suggestion
You must have [Develocity](https://gradle.com/develocity/) in order to use the Develocity Build Validation Scripts. At a minimum, your Develocity user must have the permissions `Publish build scans` and `View build scans and build data` granted explicitly or anonymously, as analyzing the results relies on the use of build scans. To get the full benefit of the Develocity Build Validation Scripts, your Develocity user should also be granted the `Access build data via the API` permission.
```

I think it's okay to mention the API permission here as well. It increases the probability that the reader will actually see it.

erichaagdev: If we call out API as a requirement then I feel we should do the same for CCUD. Both are optional. Do you think it should also be included here under "requirements"? It also has its own section already later down in the README.

jthurne: > If we call out API as a requirement then I feel we should do the same for CCUD.

Good call. I think that's a good idea as well. Then there is just one place where the user can a list of everything they may need.

The sections further down go into more detail, so maybe we can briefly mention the API permission and CCUD here, and link to the later sections.

etiennestuder, line 11: How come View build scans and build data is required?

etiennestuder: I understand that you need that permission to view / compare build scans, but strictly speaking, they are not required to use the BVS, e.g., if all you care is whether there were cache misses.

etiennestuder, line 11: Running the scripts with a user that does not have API access permissions makes the BVS of limited value.

etiennestuder, line 11: Given my comments above, I suggest rewriting this section and being explicit about what each permission is required for.

etiennestuder (changes requested):

Gradle.md
erichaagdev: I see your point.

You must have one of: `View build scans and build data` or `Access build data via the API`

Having neither and running the scripts would be pointless. I'll try to clarify all this.

SimonMarquis:
👋 Hi there, I'm the author of #777.
I don't agree with you that having a Develocity instance **is a requirement**.

Running these scripts without owning a Develocity endpoint is still very valuable. I get an easy way of detecting build cache and configuration cache issues on CI (without having to rewrite, or copy your gha). I have access to *Investigation Quick Links*, which can help me find areas of improvement. etc.

I understand that some information can be extracted without having API access (like serialization factor, etc.) but that does not mean Develocity is a requirement for these scripts. 

I would personaly be OK with mentioning that only a limited set of features are available without a Develocity endpoint, but not that it is required.

erichaagdev:
Hi, @SimonMarquis and thank you for your feedback. 

For now, this is the stance we want to take as we don't want to imply that we have any support for using `scans.gradle.com` at all. This isn't because we don't _want_ to support it, but because we simply don't have the capacity to support it at this time. That's not to say we won't have capacity to add proper support at some point in the future. I've created a ticket for this https://github.com/gradle/develocity-build-validation-scripts/issues/795, but I can't promise that it will get prioritized. 

The other option was to actively prevent the use of `scans.gradle.com`. After an internal discussion, updating the README felt like the best option for now. Doing the former would mean we would be intentionally taking away your ability to run these with `scans.gradle.com` if you choose to do so, despite our saying Develocity is a requirement.

Gradle.md
```diff
@@ -8,9 +8,47 @@ There are currently five experiments for Gradle. You could also perform these ex
...
+
+Experiment builds running locally require different permissions than those that run in CI.
+Note that while not all permissions are required, they're still strongly recommended for the best possible experience.
+
+Your Develocity user should be granted the following permissions:
+
+| User permission                 | Reason                                                                                                                                      | Required              |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|-----------------------|
+| Publish build scans             | Analyzing the experiment results relies on build scans published to Develocity.                                                             | Yes                   |
+| Read build cache data           | Reading from the remote build cache.                                                                                                        | Only for experiment 5 |
```
etiennestuder, line 30: Exp 4 (ci/ci), too. No?

etiennestuder: I assume you are referring to running locally only.

etiennestuder: Ha, it is clarified further down. Maybe split local vs. CI into subtitles?

Gradle.md
```diff
@@ -8,9 +8,47 @@ There are currently five experiments for Gradle. You could also perform these ex
...
+
+| User permission                 | Reason                                                                                                                                      | Required                                        |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------|
+| Publish build scans             | Analyzing the experiment results relies on build scans published to Develocity.                                                             | Yes                                             |
+| Read and write build cache data | Writing to the remote build cache.                                                                                                          | Yes                                             |
+| Access build data via the API   | Used to retrieve the summarized experiment results. See [Authenticating with Develocity](#authenticating-with-develocity) for more details. | Only when using `--fail-if-not-fully-cacheable` |
```
etiennestuder, line 44: I find the last sentence a bit 'confusing'.

etiennestuder (changes requested):

Gradle.md
erichaagdev: > Maybe split local vs. CI into subtitles?

Can you clarify what you mean?

Gradle.md
erichaagdev: Which sentence specifically do you find confusing?

Gradle.md
etiennestuder: Under `### User permissions`, have `#### Running the scripts locally` and `#### Running the scripts on CI`.

Gradle.md
etiennestuder: `Only when using '--fail-if-not-fully-cacheable'`. I would at least repeat that the permission is mandatory when running with the mentioned flag.

Gradle.md
```diff
@@ -19,29 +19,29 @@ Note how to deal with the eventuality of facing filename too long errors [here](
...
-| Access build data via the API   | Used to retrieve the summarized experiment results. See [Authenticating with Develocity](#authenticating-with-develocity) for more details. | No                    |
+| User permission                 | Reason                                                                                                                                      |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
+| Publish build scans             | Analyzing the experiment results relies on build scans published to Develocity.                                                             |
+| Read build cache data           | Reading from the remote build cache.                                                                                                        |
+| View build scans and build data | Build scans can be used to perform a deeper analysis of the experiment results.                                                             |
```
etiennestuder, line 31: `Performing a deeper analysis of the experiment results in Develocity.`

Gradle.md
```diff
@@ -19,29 +19,29 @@ Note how to deal with the eventuality of facing filename too long errors [here](
...
+| Access build data via the API   | Used to retrieve the summarized experiment results. See [Authenticating with Develocity](#authenticating-with-develocity) for more details. |
```
etiennestuder, line 32: `Retrieving the published build data from Develocity to summarize the experiment results.`

Gradle.md
```diff
@@ -19,29 +19,29 @@ Note how to deal with the eventuality of facing filename too long errors [here](
...
-| Access build data via the API   | Used to retrieve the summarized experiment results. See [Authenticating with Develocity](#authenticating-with-develocity) for more details. | Only when using `--fail-if-not-fully-cacheable` |
+| User permission                 | Reason                                                                                                                                      |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
+| Publish build scans             | Analyzing the experiment results relies on build scans published to Develocity.                                                             |
+| Read and write build cache data | Writing to the remote build cache.                                                                                                          |
+| Access build data via the API   | Used to retrieve the summarized experiment results. See [Authenticating with Develocity](#authenticating-with-develocity) for more details. |
```
etiennestuder, line 44: Same comments for the CI case as for the local developer case.

etiennestuder (changes requested):

Issue #287, opened 2023-01-18
Merge `mapping.example` and `network.settings` to one generalized settings file


erichaagdev:
We should merge these two files to one common settings file (e.g. `config.properties`/`settings.properties`). This reduces the number of files and provides us with a file to add more generalized configuration options.

etiennestuder:
Some thoughts:
- should we call the default file `default.config`? (one could then also have additional files like spring.config, etc.)
- I would use `-z` for the cmd line param
- we also need to update the `--help` output when we change from -m to -z
- we also need to update the cmd we print at the very end in interactive mode
- we should debate whether or not to include the config file path in the summary (the only thing not directly related to the build)
-

jthurne:
👍 Description updated accordingly.

erichaagdev:
The availability wait timeout for Develocity API calls would be a nice configuration option to have. 

https://github.com/gradle/develocity-build-validation-scripts/pull/791

PR #791, opened 2025-01-27
Scripts wait longer for builds to ingest when `--fail-if-not-fully-cacheable` is used

### LINKED ISSUE #505: Scripts do not wait long enough for results when running with `--fail-if-not-fully-cacheable`
The `availabilityWaitTimeoutSecs` of the API is always configured to be 10 seconds with no way to override it. After this 10 seconds, if there is still no build scan data available, the scripts will proceed without the performance characteristics populated. In most cases this is fine because you can use the corresponding build scan quick links to investigate.

However, having the API call succeed and the corresponding build data to be available to the scripts is critical when running with `--fail-if-no-fully-cacheable`. I propose we increase `availabilityWaitTimeoutSecs` when running with `--fail-if-not-fully-cacheable` (e.g. 30s, 60s, 120s). 

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/b5e54b97633dd851bb59b135538c722a01c7c5db/components/fetch-build-scan-data-cmdline-tool/src/main/java/com/gradle/enterprise/api/client/GradleEnterpriseApiClient.java#L42

This is related to but does not replace #493. This will ensure we get a response in more cases while #493 will fail the build if no response is received.

erichaagdev:
This change adds the `--availability-wait-timeout` option to the build scan summary tool invocation with a value of '60' when the `--fail-if-not-fully-cacheable` option is used.

This should increase the likelihood of experiment build scans being ingested in order to retrieve the required data from the Develocity API to determine cacheability.

The `availabilityWaitTimeoutSecs` query parameter is included on the Develocity API calls when the `--fail-if-not-fully-cacheable` option is included in the script invocation. 
image

When the `--fail-if-not-fully-cacheable` option isn't included, no `availabilityWaitTimeoutSecs` parameter is included in the Develocity API call.
image

Closes #505

### DIFF
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,6 +1,7 @@
 > [!IMPORTANT]
 > The distributions of the Develocity Build Validation Scripts prefixed with `gradle-enterprise` are deprecated and will be removed in a future release. Migrate to the distributions prefixed with `develocity` instead.
 
+- [FIX] Scripts do not wait long enough for build scans to become available when `--fail-if-not-fully-cacheable` is used
 - [FIX] Successful exit code returned when performance characteristics are unknown and `--fail-if-not-fully-cacheable` is used
 - [FIX] Gradle experiments do not disable background Build Scan publication
 - [FIX] Common Custom User Data Gradle plugin not injected for Gradle builds
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -61,7 +61,7 @@ val isDevelopmentRelease = !hasProperty("finalRelease")
 val releaseVersion = releaseVersion()
 val releaseNotes = releaseNotes()
 val distributionVersion = distributionVersion()
-val buildScanSummaryVersion = "1.0.1-2024.1"
+val buildScanSummaryVersion = "1.0.2-2024.1"
 
 allprojects {
     version = releaseVersion.get()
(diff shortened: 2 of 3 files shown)


gabrielfeo (approved):

release/changes.md
```diff
@@ -1,4 +1,5 @@
 > [!IMPORTANT]
 > The distributions of the Develocity Build Validation Scripts prefixed with `gradle-enterprise` are deprecated and will be removed in a future release. Migrate to the distributions prefixed with `develocity` instead.
 
+- [NEW] Wait up to 60 seconds for experiment build scans to become available when `--fail-if-not-fully-cacheable` is used
```
ribafish, line 4: Would it make sense to expose this as a user-configurable setting somehow, either via commandline, or with the `network.settings` file?

ribafish (approved):

release/changes.md
erichaagdev: Yes, and we would like to, but not yet. We have this issue #287 to create a generalized settings file for settings like this one, but it will introduce breaking changes so we want to wait until the next major version to implement it. 

Until then, this should suffice given the value already isn't configurable.

jthurne (approved):

PR #794, opened 2025-02-03
Exception is written to `exceptions.txt` file on build scan error

erichaagdev:
This change adds a new file `exceptions.txt` that writes the contents of the `error` payload to a file.


components/scripts/gradle/gradle-init-scripts/configure-build-validation.gradle
```diff
@@ -54,6 +54,8 @@ def registerBuildScanActions = { def buildScan, def rootProjectName ->
     buildScan.onError { error ->
         def errorFile = new File(expDir, 'errors.txt')
         errorFile.text = 'Build Scan publishing failed.'
+        def exceptionFile = new File(expDir, 'exceptions.txt')
+        exceptionFile.text = error
```
jthurne, line 58: It took me a moment to look up what `error` is.  It looks like it's a Strong containing a JSON object that describes the error. So I think writing out the error variable like this is sufficient.

jthurne (approved):

PR #807, opened 2025-02-27
Develocity and Gradle Enterprise remote build cache connectors are supported

erichaagdev:
> [!NOTE]
> Draft only because the description is a WIP. The changes are otherwise ready for review. 

## Overview

Adds a new command line argument for Gradle experiment 5, `-y` or `--remote-build-cache-type`, allowing the user to specify a remote build cache type to use. When specified, valid values are: `develocity`, `gradle-enterprise`, or `http`. When the argument is specified, it will always take precedence over what is configured in the build.

If the argument is not specified, the remote build cache type of the first build will be fetched from the Develocity API and used for the second build.

If the argument is not specified and the user does not have access to the Develocity API, then the existing remote build cache configuration will be used.

### New `-y` / `--remote-build-cache-type` argument

A new `-y` / `--remote-build-cache-type` argument was added for Gradle experiment 5 which is visible in the help text.

image

Valid values are: `http`, `gradle-enterprise` and `develocity`. If an invalid value is supplied, the script will immediately fail.

image

When running in interactive mode, the user will be prompted to enter a value. The below image defaulted to `develocity` since that's what the first build used (fetched from the Develocity API). 

image

### Remote build cache type from Develocity API

If `--remote-build-cache-type` is not specified, the remote build cache type of the first build will be fetched from the Develocity API and used for the second build. The `--remote-build-cache-type` always takes precedence.

### DIFF
--- a/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
+++ b/components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
@@ -40,6 +40,7 @@ develocity_server=''
 interactive_mode=''
 
 ci_build_scan_url=''
+remote_build_cache_type=''
 remote_build_cache_url=''
 mapping_file=''
 
@@ -110,9 +111,9 @@ wizard_execute() {
   collect_gradle_details
 
   print_bl
-  explain_remote_build_cache_url
+  explain_collect_remote_build_cache
   print_bl
-  collect_remote_build_cache_url
+  collect_remote_build_cache
   explain_command_to_repeat_experiment_after_collecting_parameters
 
   print_bl
@@ -136,6 +137,7 @@ wizard_execute() {
 
 map_additional_script_args() {
   ci_build_scan_url="${_arg_first_build_ci}"
+  remote_build_cache_type="${_arg_remote_build_cache_type}"
   remote_build_cache_url="${_arg_remote_build_cache_url}"
   mapping_file="${_arg_mapping_file}"
 }
@@ -151,6 +153,10 @@ validate_required_args() {
   if [[ "${enable_develocity}" == "on" && -z "${develocity_server}" ]]; then
     _PRINT_HELP=yes die "ERROR: Missing required argument when enabling Develocity on a project not already connected: --develocity-server" "${INVALID_INPUT}"
   fi
+
+  if [[ -n "${remote_build_cache_type}" && "${remote_build_cache_type}" != 'http' && "${remote_build_cache_type}" != 'gradle-enterprise' && "${remote_build_cache_type}" != 'develocity' ]]; then
+    _PRINT_HELP=yes die "ERROR: Invalid value for argument --remote-build-cache-type. Must be one of: 'http', 'gradle-enterprise', or 'develocity'" "${INVALID_INPUT}"
+  fi
 }
 
 fetch_build_params_from_build_scan() {
@@ -160,6 +166,9 @@ fetch_build_params_from_build_scan() {
 }
 
 read_build_params_from_build_scan_data() {
+  if [[ "${remote_build_cache_types[0]}" == "disabled" ]]; then
+    die "ERROR: Remote build cache was disabled for the first build. Enable the remote build cache in the build and restart the experiment."
+  fi
   if [ -z "${git_repo}" ]; then
     git_repo="${git_repos[0]}"
     project_name="$(basename -s .git "${git_repo}")"
@@ -169,6 +178,9 @@ read_build_params_from_build_scan_data() {
   fi
   if [ -z "${git_commit_id}" ]; then
     git_commit_id="${git_commit_ids[0]}"
+  fi
+  if [[ -z "${remote_build_cache_type}" && "${remote_build_cache_types[0]}" != "unknown" ]]; then
+    remote_build_cache_type="${remote_build_cache_types[0]}"
   fi
   if [ -z "${remote_build_cache_url}" ]; then
     remote_build_cache_url="${remote_build_cache_urls[0]}"
@@ -196,6 +208,9 @@ validate_build_config() {
 execute_build() {
   local args
   args=(--build-cache --init-script "${INIT_SCRIPTS_DIR}/configure-remote-build-caching.gradle")
+  if [ -n "${remote_build_cache_type}" ]; then
+    args+=("-Ddevelocity.build-validation.remoteBuildCacheType=${remote_build_cache_type}")
+  fi
   if [ -n "${remote_build_cache_url}" ]; then
     args+=("-Ddevelocity.build-validation.remoteBuildCacheUrl=${remote_build_cache_url}")
   fi
@@ -360,7 +375,7 @@ EOF
   print_interactive_text "${text}"
 }
 
-explain_remote_build_cache_url() {
+explain_collect_remote_build_cache() {
   local text
   IFS='' read -r -d '' text <<EOF
 The local build will connect to the given remote build cache. The remote build
@@ -369,11 +384,30 @@ EOF
   print_interactive_text "${text}"
 }
 
-collect_remote_build_cache_url() {
-  local default_remote_cache="<project default>"
-  prompt_for_setting "What is the remote build cache url to use?" "${remote_build_cache_url}" "${default_remote_cache}" remote_build_cache_url
+collect_remote_build_cache() {
+  collect_remote_build_cache_type
+  collect_remote_build_cache_url
+}
 
-  if [[ "${remote_build_cache_url}" == "${default_remote_cache}" ]]; then
+collect_remote_build_cache_type() {
+  local default_remote_cache_type="<project default>"
+  prompt_for_setting "What is the remote build cache type to use? Must be: 'http', 'gradle-enterprise', 'develocity', or blank for project default." "${remote_build_cache_type}" "${default_remote_cache_type}" remote_build_cache_type
+
+  if [[ -n "${remote_build_cache_type}" && "${remote_build_cache_type}" != 'http' && "${remote_build_cache_type}" != 'gradle-enterprise' && "${remote_build_cache_type}" != 'develocity' ]]; then
+    print_bl
+    die "ERROR: Invalid value for remote build cache type. Must be one of: 'http', 'gradle-enterprise', 'develocity', or blank for project default" "${INVALID_INPUT}"
+  fi
+
+  if [[ "${remote_build_cache_type}" == "${default_remote_cache_type}" ]]; then
+    remote_build_cache_type=''
+  fi
+}
+
+collect_remote_build_cache_url() {
+  local default_remote_cache_url="<project default>"
+  prompt_for_setting "What is the remote build cache url to use?" "${remote_build_cache_url}" "${default_remote_cache_url}" remote_build_cache_url
+
+  if [[ "${remote_build_cache_url}" == "${default_remote_cache_url}" ]]; then
     remote_build_cache_url=''
   fi
 }
@@ -463,6 +497,10 @@ generate_command_to_repeat_experiment() {
 
   if [ -n "${mapping_file}" ]; then
     cmd+=("-m" "${mapping_file}")
+  fi
+
+  if [ -n "${remote_build_cache_type}" ]; then
+    cmd+=("-y" "${remote_build_cache_type}")
   fi
 
   if [ -n "${remote_build_cache_url}" ]; then
--- a/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
+++ b/components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
@@ -4,23 +4,109 @@ static getInputParam(Gradle gradle, String name) {
     return gradle.startParameter.systemPropertiesArgs[name] ?: System.getProperty(name) ?: System.getenv(envVarName)
 }
 
+def isTopLevelBuild = !gradle.parent
+
+def expDir = getInputParam(gradle, 'develocity.build-validation.expDir')
+def remoteBuildCacheType = getInputParam(gradle, 'develocity.build-validation.remoteBuildCacheType')
 def remoteBuildCacheUrl = getInputParam(gradle, 'develocity.build-validation.remoteBuildCacheUrl')
 
-settingsEvaluated { settings ->
+def docsRoot = 'https://docs.gradle.com/develocity/gradle-plugin'
+
+settingsEvaluated { Settings settings ->
     settings.buildCache {
         local {
             enabled = false
         }
-        remote(HttpBuildCache) {
-            enabled = true
-            push = false
-            if (remoteBuildCacheUrl) {
-                url = withPathTrailingSlash(new URI(remoteBuildCacheUrl))
+
+        if (remoteBuildCacheType) {
+            if (!['develocity', 'gradle-enterprise', 'http'].contains(remoteBuildCacheType)) {
+                // The scripts already fail if the value of --remote-build-cache-type isn't valid.
+                // This is for the sake of completeness since this init script assumes it's valid.
+                def errorFile = new File(expDir, 'errors.txt')
+                errorFile.text = "Unknown remote build cache type: ${remoteBuildCacheType}"
+                throw new IllegalStateException("Unknown remote build cache type: ${remoteBuildCacheType}")
             }
+
+            if (isTopLevelBuild && missingRequiredPlugin(settings, remoteBuildCacheType)) {
+                // Included builds may not have the necessary plugin applied.
+                // Only fail if the top-level build is missing the required extension.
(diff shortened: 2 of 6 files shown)


components/scripts/lib/cli-parsers/gradle/05-cli-parser.m4
```diff
@@ -30,6 +31,7 @@ function print_help() {
   print_option_usage -p
   print_option_usage -t
   print_option_usage -a
+  print_option_usage "-y, --remote-build-cache-type" "Specifies the type of remote build cache to use in the second build run locally. Must be one of: 'http', 'gradle-enterprise', or 'develocity'"
```
erichaagdev, line 34: @etiennestuder I recently went through the scripts and removed all public facing usages of Gradle Enterprise. These changes now add some back. 

Something to consider is to not show `gradle-enterprise` in anything public facing, but still accept it as a value and still configure it. This is exactly what we did for the options like `--enable-gradle-enterprise`. 

On one hand it feels a bit unfortunate to regress on the "no public facing usages of GE", but on the other it feels bad implementing support for something that is immediately deprecated. 

We also need to determine how this functionality will work with 3.0. Will we still accept `gradle-enterprise` as a value?

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -160,6 +166,9 @@ fetch_build_params_from_build_scan() {
 }
 
 read_build_params_from_build_scan_data() {
+  if [[ "${remote_build_cache_types[0]}" == "disabled" ]]; then
+    die "ERROR: Remote build cache was disabled for the first build. Enable the remote build cache in the build and restart the experiment."
```
erichaagdev, line 170: This does prevent users from doing something they could previously which is to invoke the script for a build that didn't have the remote build cache enabled, but I can't think of a valid use case for that.

The benefit to the user is that we will quickly catch what is likely a mistake (e.g., using the wrong Build Scan) and this happens before the first build is even invoked.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -369,11 +384,30 @@ EOF
...
+}
+
+collect_remote_build_cache_type() {
+  local default_remote_cache_type="<project default>"
+  prompt_for_setting "What is the remote build cache type to use? Must be: 'http', 'gradle-enterprise', 'develocity', or blank for project default." "${remote_build_cache_type}" "${default_remote_cache_type}" remote_build_cache_type
+
+  if [[ -n "${remote_build_cache_type}" && "${remote_build_cache_type}" != 'http' && "${remote_build_cache_type}" != 'gradle-enterprise' && "${remote_build_cache_type}" != 'develocity' ]]; then
+    print_bl
+    die "ERROR: Invalid value for remote build cache type. Must be one of: 'http', 'gradle-enterprise', 'develocity', or blank for project default" "${INVALID_INPUT}"
+  fi
```
erichaagdev, line 399: This is the first option for which there are valid values. Failing on an invalid value was easiest to implement, but looping until we get a valid value would provide a better user experience. Let me know what you think.

components/scripts/lib/summary.sh
```diff
@@ -119,20 +119,34 @@ detect_warnings_from_build_scans() {
...
 
   if [[ "${value_mismatch}" == "true" ]]; then
     warnings+=("Differences were detected between the two builds. This may skew the outcome of the experiment.")
   fi
+  if [ "${remote_build_cache_urls[0]}" != "${remote_build_cache_urls[1]}" ]; then
+    warnings+=("The two builds ran with different remote build cache URLs configured.")
+  fi
+  if [ "${remote_build_cache_class_names[0]}" != "${remote_build_cache_class_names[1]}" ]; then
+    warnings+=("The two builds ran using different remote build cache implementations.")
+  fi
```
erichaagdev, line 149: It's worth noting that in these scenarios, the user will get at least 2 warning messages. 

```
First: 

- Differences were detected between the two builds. This may skew the outcome of the experiment.

Followed by one or both of: 

- The two builds ran with different remote build cache URLs configured.
- The two builds ran using different remote build cache implementations.
```

No other mismatch gets an explicit callout, but I felt it made sense here because we don't show the remote build cache URL or type in the _Summary_ section.

The other option is to add remote build cache URL or type to the _Summary_ section.

We also don't need to also print `Differences were detected between the two builds. This may skew the outcome of the experiment.`, but I like including `This may skew the outcome of the experiment` because it's true.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
gabrielfeo: IMHO, it's fine to simply fail because valid values are clearly stated and there's not much room for typos.

components/scripts/lib/summary.sh
gabrielfeo: Adding the values to the Summary and removing their individual warnings in favor of the existing one would make the most sense to me, because then the behavior would be in line with other cases. For example, when the builds ran different tasks: values compared side-by-side in the summary + a single warning for the entire Summary section.

Screenshot 2025-03-03 at 13 37 16

gabrielfeo (commented):
Awesome work! This will improve the experience for many users

components/scripts/lib/summary.sh
erichaagdev: It is possible to do it that way. 👍

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -369,11 +384,30 @@ EOF
...
 }
 
+collect_remote_build_cache() {
+  collect_remote_build_cache_type
+  collect_remote_build_cache_url
+}
+
+collect_remote_build_cache_type() {
+  local default_remote_cache_type="<project default>"
+  prompt_for_setting "What is the remote build cache type to use? Must be: 'http', 'gradle-enterprise', 'develocity', or blank for project default." "${remote_build_cache_type}" "${default_remote_cache_type}" remote_build_cache_type
```
erichaagdev, line 394:
```suggestion
  prompt_for_setting "What is the remote build cache type to use? [http, gradle-enterprise, develocity, <BLANK>]" "${remote_build_cache_type}" "${default_remote_cache_type}" remote_build_cache_type
```

components/scripts/lib/summary.sh
```diff
@@ -119,20 +119,34 @@ detect_warnings_from_build_scans() {
     if [ -z "${build_outcomes[i]}" ]; then
       warnings+=("Failed to fetch build scan data for the ${ORDINALS[i]} build.")
     fi
+    if [ "${remote_build_cache_types[i]}" == "unknown" ]; then
+      # "Develocity Build Validation Scripts" is specifically mentioned as to
+      # not imply Develocity itself does not work with other remote build cache
+      # implementations.
+      warnings+=("The ${ORDINALS[i]} build ran using a remote build cache implementation not officially supported by the Develocity Build Validation Scripts.")
```
erichaagdev, line 126:
```suggestion
      warnings+=("The ${ORDINALS[i]} build ran using a remote build cache implementation not officially supported by the experiment.")
```

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -369,11 +384,30 @@ EOF
...
+
+  if [[ -n "${remote_build_cache_type}" && "${remote_build_cache_type}" != 'http' && "${remote_build_cache_type}" != 'gradle-enterprise' && "${remote_build_cache_type}" != 'develocity' ]]; then
+    print_bl
+    die "ERROR: Invalid value for remote build cache type. Must be one of: 'http', 'gradle-enterprise', 'develocity', or blank for project default" "${INVALID_INPUT}"
```
erichaagdev, line 398:
```suggestion
    die "ERROR: Invalid value for remote build cache type. Must be one of: http, gradle-enterprise, develocity, or <BLANK> for project default" "${INVALID_INPUT}"
```

erichaagdev: Is there `.` a dot?

components/scripts/lib/cli-parsers/gradle/05-cli-parser.m4
erichaagdev: `develocity`, `gradle-enterprise`, `http`

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
```diff
@@ -4,23 +4,109 @@ static getInputParam(Gradle gradle, String name) {
...
+            }
+
+            def remoteBuildCacheImplementation = getRemoteBuildCacheImplementation(settings, remoteBuildCacheType)
+            if (remoteBuildCacheImplementation) {
+                logger.debug("Configuring remote build cache type for '${settings.rootProject.name}' as: ${remoteBuildCacheImplementation}")
+                remote(remoteBuildCacheImplementation)
             }
         }
+
+        logger.debug("Remote build cache type for '${settings.rootProject.name}' is: ${remote?.class?.name}")
```
erichaagdev, line 55:
```suggestion
        logger.debug("Remote build cache implementation for '${settings.rootProject.name}' is: ${remote?.class?.name}")
```

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
```diff
@@ -4,23 +4,109 @@ static getInputParam(Gradle gradle, String name) {
...
+                        "when running the build validation script.")
+                } else {
+                    throw new IllegalStateException("Remote build cache type 'gradle-enterprise' requested,\n" +
+                        "but the Gradle Enterprise Gradle plugin is not applied (see $docsRoot/legacy/#applying_the_plugin).")
+                }
+            }
+
+            def remoteBuildCacheImplementation = getRemoteBuildCacheImplementation(settings, remoteBuildCacheType)
+            if (remoteBuildCacheImplementation) {
+                logger.debug("Configuring remote build cache type for '${settings.rootProject.name}' as: ${remoteBuildCacheImplementation}")
```
erichaagdev, line 50:
```suggestion
                logger.debug("Configuring remote build cache implementation for '${settings.rootProject.name}' as: ${remoteBuildCacheImplementation}")
```

components/scripts/lib/summary.sh
erichaagdev: We decided to keep it how I have it implemented now.

components/scripts/lib/cli-parsers/gradle/05-cli-parser.m4
erichaagdev: We decided to leave it. We can discuss what to do for 3.0 when the time comes.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
jthurne: >  I can't think of a valid use case for that

What if the user is using the BVS to test or explore something other than build caching?

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -369,11 +384,30 @@ EOF
...
 }
 
+collect_remote_build_cache() {
+  collect_remote_build_cache_type
+  collect_remote_build_cache_url
+}
+
+collect_remote_build_cache_type() {
+  local default_remote_cache_type="<project default>"
+  prompt_for_setting "What is the remote build cache type to use? [develocity, gradle-enterprise, http, or <BLANK>]" "${remote_build_cache_type}" "${default_remote_cache_type}" remote_build_cache_type
```
jthurne, line 394: The Gradle documentation and the Develocity Gradle plugin end-user documentation calls these "Build cache connectors".  I wonder if we should replace "type" with "connector" so that the BVS are consistent with the docs.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
```diff
@@ -369,11 +384,30 @@ EOF
...
+
+  if [[ -n "${remote_build_cache_type}" && "${remote_build_cache_type}" != 'http' && "${remote_build_cache_type}" != 'gradle-enterprise' && "${remote_build_cache_type}" != 'develocity' ]]; then
+    print_bl
+    die "ERROR: Invalid value for remote build cache type. Values are 'develocity', 'gradle-enterprise', 'http', or <BLANK> for project default." "${INVALID_INPUT}"
```
jthurne, line 398: It is not very user-friendly to completely exit the script in interactive mode if the user enters an invalid value. It would be better to print the error message, and call `collect_remote_build_cache_type()` again (or do the validation   using a do...while loop) until the user does provide a valid value.

jthurne: I also wonder if we should include a link to some documentation in the error message so that users who don't know what this is can learn more about it.

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
```diff
@@ -4,23 +4,110 @@ static getInputParam(Gradle gradle, String name) {
...
+def isTopLevelBuild = !gradle.parent
+
+def expDir = getInputParam(gradle, 'develocity.build-validation.expDir')
+def remoteBuildCacheType = getInputParam(gradle, 'develocity.build-validation.remoteBuildCacheType')
 def remoteBuildCacheUrl = getInputParam(gradle, 'develocity.build-validation.remoteBuildCacheUrl')
 
-settingsEvaluated { settings ->
+def docsRoot = 'https://docs.gradle.com/develocity/gradle-plugin'
+
+settingsEvaluated { Settings settings ->
```
jthurne, line 15: There is now a lot of code in this block, and a lot of nested if statements. I think it would be helpful to extract some of the nested if statements into helper functions so that the code is overall easier to read.

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
```diff
@@ -4,23 +4,110 @@ static getInputParam(Gradle gradle, String name) {
...
+
+        if (remoteBuildCacheType) {
+            if (!['develocity', 'gradle-enterprise', 'http'].contains(remoteBuildCacheType)) {
+                // The scripts already fail if the value of --remote-build-cache-type isn't valid.
+                // This is for the sake of completeness since this init script assumes it's valid.
+                def errorFile = new File(expDir, 'errors.txt')
+                def message = "Invalid value '${remoteBuildCacheType}' for remote build cache type. Values are 'develocity', 'gradle-enterprise', or 'http'."
+                errorFile.text = message
+                throw new IllegalStateException(message)
+            }
```
jthurne, line 29: >  There is now a lot of code in this block, and a lot of nested if statements. I think it would be helpful to extract some of the nested if statements into helper functions so that the code is overall easier to read.

For example, this code block could be extracted into a `validateRemoteBuildCacheType` function.

jthurne (changes requested):

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
erichaagdev: > do the validation using a do...while loop

@etiennestuder and I discussed exactly this and decided to move forward with just exiting for now. It comes at the cost of user-friendliness, but it's the most straight forward approach. If it turns out to be problematic for users we can certainly fix it.

> I also wonder if we should include a link to some documentation in the error message so that users who don't know what this is can learn more about it.

That's not a bad idea, but would require documentation to link to. 😅 

I've created issue https://github.com/gradle/develocity-build-validation-scripts/issues/809 for this so that it can be addressed in a subsequent PR.

components/scripts/gradle/05-validate-remote-build-caching-ci-local.sh
etiennestuder: It's a good point. 'connector' would be the correct term. Also, build cache type sounds like the type of the build cache -- and not like the class type used to coonect to the remote build cache.

components/scripts/gradle/gradle-init-scripts/configure-remote-build-caching.gradle
erichaagdev: @jthurne I didn't necessarily address the nesting, but I significantly reduced the cognitive overload by extracting the error handling. The core logic first entirely on my screen which was not the case before. Do you think it's sufficient? See a8b7e1e693f000bcc17ad4eeef67c475c15e2ba8

jthurne: That is a significant improvement 👍

PR #813, opened 2025-03-27
Experiment `run_id` is made significantly more random

### LINKED ISSUE #496: Experiment Run ID should be more unique
Recently we had a collision on "experiment run id" for three automated experiment runs. We should reconsider how we determine this value. Currently it's computed as the hex representation of the current timestamp seconds.

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/fc1a6d6fc6e87c22ceb3d4f2c4fa50d8ea9a2fa9/components/scripts/lib/init.sh#L17-L19

At first it seems seems unlikely that two or more experiments will compute the same run id, but given that we now have 39 (at the time of this writing) projects running automated experiments, each running 2-3 experiments, and them all starting at the exact same time Sunday @ 9:00 UTC, it's bound to happen often.

Example: https://ge.solutions-team.gradle.com/scans?search.names=Experiment%20run%20id&search.timeZoneId=America/Chicago&search.values=64f44be9&selection.buildScanB=ygdfpgm4geobu

erichaagdev:
See issue #496 for the reason _why_ this was done.

For an explanation of _how_ it was done, see the code comment above the `generate_run_id` function that explains this in great detail. 

Example: https://ge.solutions-team.gradle.com/scans?search.names=Experiment%20run%20id&search.timeZoneId=America%2FChicago&search.values=1c87416e774b51d#selection.buildScanA=pn6cdtogklovg

Resolves #496

### DIFF
--- a/components/scripts/lib/init.sh
+++ b/components/scripts/lib/init.sh
@@ -14,8 +14,33 @@ make_symlink_to_latest_experiment_dir() {
   ln -s "${EXP_DIR}" "${LINK_NAME}" > /dev/null 2>&1
 }
 
+# Below is an explanation of what constitutes a 'run_id'.
+#
+#    0:     A static digit '1'
+#    1-5:   A random number, left-padded with '0', then reversed
+#    6-16:  The current number of seconds since epoch, truncated to 10 digits, then reversed
+#    17-18: The experiment number
+#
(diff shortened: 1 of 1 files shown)


jthurne (approved):
Nice!

jthurne (approved):

PR #817, opened 2025-03-28
Use `--experiment-run-id` argument when calling summary tool

erichaagdev:
This PR updates the scripts to include the `--experiment-run-id` argument to the build scan summary tool when both build scans are published by the scripts. This makes it possible for the tool to fetch all build scan data at once. For this to work, both builds must have a common experiment run id. 

image

In scenarios where there is no common run id, as is the case in experiments involving CI builds, only 2 calls are made. 

image

Previously, a separate call was made for each required build model.

image


components/scripts/lib/build-scan-online.sh
erichaagdev, line 1: Which experiments called what function in this file and with what arguments was confusing to understand. Some called `process_build_scan_data_online` and others called `fetch_build_scans_and_build_time_metrics` directly.

While slightly more duplication, I feel I've made it much simpler by removing `fetch_build_scans_and_build_time_metrics` and having all every script call `process_build_scan_data_online` and explicitly set the log level and desired query.

jthurne (approved):

jthurne (approved):

erichaagdev:
Here's a reference build scan with the removed third party dependencies. https://ge.solutions-team.gradle.com/s/3cm4thmi33bkg/dependencies

PR #821, opened 2025-04-10
Expose GitHub summary unconditionally

jprinet:
### Issue
The GitHub summary is not published when the validation scripts run as part of the [GiHub composite actions](https://github.com/gradle/develocity-build-validation-scripts/tree/main/.github/actions) fail.

### Fix 
The validation scripts are run within a bash step which is by default configured to exit immediately upon command failure
```
set -e
```
The fix is to revert this configuration and capture the validation script exit code to allow post actions within the step (build scan links capture and receipt file renaming)
Subsequent steps to archive the receipt and inline it in the summary are configured to run unconditionnally
```
if: always()
```

### Implementation details
- The receipt file is renamed to  a predictable receipt.txt file to avoid passing the filename across steps and adding an attack surface as the file content is archived and inlined in the GitHub summary
- the `upload-artifact` step is using the default configuration which triggers a warning message when the receipt file is not present. This is an interesting information for the action consumer.

### DIFF
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -122,12 +122,6 @@ runs:
         echo "buildScanFirstBuild=$BUILD_SCAN_1" >> $GITHUB_OUTPUT
         echo "buildScanSecondBuild=$BUILD_SCAN_2" >> $GITHUB_OUTPUT
         
-        # Copy receipt to a predictable location
-        RECEIPT_FILE=".data/01-validate-incremental-building/latest/exp1-*.receipt"
-        if [ -f ${RECEIPT_FILE} ]; then
-          cp ${RECEIPT_FILE} ../receipt.txt
-        fi
-
         exit $EXPERIMENT_EXIT_CODE
       shell: bash
     - name: Archive receipt
@@ -136,12 +130,13 @@ runs:
       if: always()
       with:
         name: experiment-1-receipt-${{ github.job }}${{ strategy.job-total > 1 && format('-{0}', strategy.job-index) || '' }}
-        path: receipt.txt
+        path: develocity-gradle-build-validation/.data/01-validate-incremental-building/latest/exp1-*.receipt
     - name: Fill GitHub summary
       if: always()
       run: |
-        if [ -f "receipt.txt" ]; then
-          cat receipt.txt >> $GITHUB_STEP_SUMMARY
+        RECEIPT_FILE="develocity-gradle-build-validation/.data/01-validate-incremental-building/latest/exp1-*.receipt"
+        if [ -f ${RECEIPT_FILE} ]; then
+          cat ${RECEIPT_FILE} >> $GITHUB_STEP_SUMMARY
           echo "-------------" >> $GITHUB_STEP_SUMMARY
           echo "Download receipt: ${{ steps.upload-artifact.outputs.artifact-url }}" >> $GITHUB_STEP_SUMMARY
         fi
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -115,7 +115,7 @@ runs:
           ${RUNNER_DEBUG:+"--debug"}
         EXPERIMENT_EXIT_CODE=$?
 
-        # Set step outputs
+        # Set the Build Scan urls as outputs
         BUILD_SCAN_1=$(grep -m 1 "first build" ${RECEIPT_FILE} | sed 's/.* //')
         BUILD_SCAN_2=$(grep -m 1 "second build" ${RECEIPT_FILE} | sed 's/.* //')
         
--- a/.github/actions/gradle/experiment-1/action.yml
+++ b/.github/actions/gradle/experiment-1/action.yml
@@ -98,6 +98,9 @@ runs:
         # Navigate into the folder containing the validation scripts
         cd develocity-gradle-build-validation
 
+        # Do not exit on error to allow post-actions
+        set +e
+
         # Run the experiment
         ./01-validate-incremental-building.sh \
           ${ARG_GIT_REPO:+"-r" "$ARG_GIT_REPO"} \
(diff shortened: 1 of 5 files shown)


.github/actions/gradle/experiment-1/action.yml
```diff
@@ -110,25 +113,36 @@ runs:
...
         
         echo "buildScanFirstBuild=$BUILD_SCAN_1" >> $GITHUB_OUTPUT
         echo "buildScanSecondBuild=$BUILD_SCAN_2" >> $GITHUB_OUTPUT
         
-        cat $RECEIPT_FILE >> $GITHUB_STEP_SUMMARY
+        # Copy receipt to a predictable location
+        RECEIPT_FILE=".data/01-validate-incremental-building/latest/exp1-*.receipt"
+        if [ -f ${RECEIPT_FILE} ]; then
+          cp ${RECEIPT_FILE} ../receipt.txt
+        fi
```
erichaagdev, line 129: What do you think about something like this?

```suggestion
        echo "receiptFile=$(readlink -f .data/01-validate-incremental-building/latest/exp1-*.receipt)" >> $GITHUB_OUTPUT
```

That way we don't have to copy files around to be used in a later step. I think that would also allow users to consume it and could write the file contents to Slack or an email or something without having to worry about where the file is located. I think that's true anyway.

Then later we can do:

```shell
if [ -f "${{ steps.run.outputs.receiptFile }}" ]; then
...
fi
```

erichaagdev (approved):
1 minor non-blocking suggestion.

.github/actions/gradle/experiment-1/action.yml
jprinet: To expose the file outside of the composite action, we would also need to declare it as an extra output [there](https://github.com/gradle/develocity-build-validation-scripts/blob/e173ee8cf2815af3b687701b566c3ce39f04653f/.github/actions/gradle/experiment-1/action.yml#L40).
We could do that but my intention is exactly opposite as described there
> The receipt file is renamed to a predictable receipt.txt file to avoid passing the filename across steps and adding an attack surface as the file content is archived and inlined in the GitHub summary

We also could keep the reference `.data/01-validate-incremental-building/latest/exp1-*.receipt` in the 3 different steps to avoid the `copy`
The `copy` is just there to avoid repeating the uneasy reference.
Do you think it would be better?

erichaagdev: > To expose the file outside of the composite action, we would also need to declare it as an extra output [there](https://github.com/gradle/develocity-build-validation-scripts/blob/e173ee8cf2815af3b687701b566c3ce39f04653f/.github/actions/gradle/experiment-1/action.yml#L40).

Right, I knew there was something I was missing. I would do that.

> adding an attack surface as the file content is archived and inlined in the GitHub summary

I don't get this part.

jprinet: Having as a step input a file path which you then display in the summary and upload as an artifact is something I am a bit reluctant to do

erichaagdev: Sorry, but I still don't understand. We already displayed the content in the summary before here.

https://github.com/gradle/develocity-build-validation-scripts/blob/159dc34d6de1b2019eb9f9807d2ff14a2b90ff1c/.github/actions/gradle/experiment-1/action.yml#L122

And the file was an input to the upload action here.

https://github.com/gradle/develocity-build-validation-scripts/blob/159dc34d6de1b2019eb9f9807d2ff14a2b90ff1c/.github/actions/gradle/experiment-1/action.yml#L129

jprinet: in both cases, the file name was configured explicitly to be `develocity-gradle-build-validation/.data/01-validate-incremental-building/latest*/exp1-*.receipt `
I don't want to create a step leaking the content of a file which is defined as a previous step output.
I can imagine a scenario where this output could be overriden with a path to a file containing secrets in the consuming workflow.
You can argue that there are other easy ways to capture the file in such conditions, but I'd rather not create a step like this

erichaagdev: The artifact is uploaded immediately after the experiment completes in the following step and the summary is written immediately after that. All of this happens within the composite action so the user can overwrite the file, but neither the uploaded artifact or experiment summary should be affected. 

In any case, they could have overwritten the file before these changes too.

jprinet: I agree that the risk is limited or not worsened but I still feel it is bad practice to display a file content without controlling how the file path is obtained

erichaagdev: Is your concern only about exposing the file path to the user, i.e., adding it as an output of the composite action?

jprinet: My concern is that the composite step is displaying a file which could potentially be something else than the report.
You could think of a scenario with a complex consuming workflow, calling this composite, with maintainers thoroughly checking that files with secrets are not undisclosed by a PR, but omitting to control the behavior of steps brought by composite actions.
This is unlikely to happen but not impossible if we make the receipt file a step input/output

jprinet: We agreed on a call to inline the full receipt path when used

.github/actions/gradle/experiment-1/action.yml
```diff
@@ -110,25 +113,31 @@ runs:
           ${ARG_DEVELOCITY_URL:+"-s" "$ARG_DEVELOCITY_URL"} \
           ${ARG_DEVELOCITY_ENABLE:+"-e"} \
           ${RUNNER_DEBUG:+"--debug"}
+        EXPERIMENT_EXIT_CODE=$?
 
         # Set the Build Scan urls as outputs
-        RECEIPT_FILE=".data/01-validate-incremental-building/latest/exp1-*.receipt"
         BUILD_SCAN_1=$(grep -m 1 "first build" ${RECEIPT_FILE} | sed 's/.* //')
```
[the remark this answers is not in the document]
jprinet: damned, I missed this one, good catch. I'll fix it in another PR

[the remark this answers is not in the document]
erichaagdev: You could just move the capturing of build scan links to the `Fill GitHub summary` step inside of the `if` condition such that the final thing done in this step is the script invocation. That would also allow you to remove the `set +e` and capturing of the exit code.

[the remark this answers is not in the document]
jprinet: That makes sense and will put altogether the GitHub I/O in the same step 👍

PR #851, opened 2025-08-01
Build is upgraded to Gradle 9.0.0

erichaagdev:
This PR upgrades the Gradle version used by the build to 9.0.0.

All changes were explicitly required for a successful build.

The previous PR #850 did not account for the changes to [file permissions in archives](https://docs.gradle.org/9.0.0/release-notes.html#archive-tasks-produce-reproducible-archives-by-default) and had to be reverted.


clayburn (approved):

build.gradle.kts
```diff
@@ -278,44 +277,37 @@ val copyLegacyMavenScripts by tasks.registering(Sync::class) {
...
-    archiveFileName.set(archiveBaseName.flatMap { a -> distributionVersion.map { v -> "$a-$v.zip" } })
-    from(copyLegacyGradleScripts)
-    into(archiveBaseName.get())
+    distributionName = "gradle-enterprise-gradle-build-validation"
+    distributionContents.from(copyLegacyGradleScripts)
 }
 
-val assembleLegacyMavenScripts by tasks.registering(Zip::class) {
+val assembleLegacyMavenScripts by tasks.registering(AssembleScripts::class) {
+    description = "Packages the Gradle experiment scripts in a zip archive."
```
gabrielfeo, line 299: I think `PackageScripts` would be a more accurate name for the class, then assemble could be a lifecycle task if it must have that name (Minor suggestion, feel free not to change)
```suggestion
val assembleLegacyMavenScripts by tasks.registering(PackageScripts::class) {
    description = "Packages the Gradle experiment scripts in a zip archive."
```

gabrielfeo (approved):

PR #926, opened 2026-03-27
Require Build Scans when failing on non-cacheable builds

ribafish:
## Summary
When `--fail-if-not-fully-cacheable` is set, the Build Scan data must be available for the summary tool to determine cacheability. Previously the fetch could fail if the data was not yet ingested on the server.

- BVS passes a new `--require-build-scans` flag to `fetch-build-scan-data` **only** in the fully-cacheable/CI mode (i.e. when `--fail-if-not-fully-cacheable` is active). This tells the summary tool to use its long default wait for Build Scan data to become available; without it, the short default applies.
- The summary tool now owns the wait-time decision. BVS no longer parses `network.settings` or computes/passes a wait value.
- A new `max.wait.time` property (ISO 8601 duration, e.g. `PT2M`) can be set in `network.settings` to override the wait time; a user-set value always wins over the mode default.

This supersedes the earlier approach (BVS reading `network.settings` and passing `--max-total-wait-time`).

## Dependencies / blockers
- Depends on the build-scan-summary change in gradle/dv#63516, which adds the `--require-build-scans` flag and moves the wait-time decision into the summary tool. The flag string `--require-build-scans` must stay in sync between the two PRs.
- **Blocked on publishing:** after gradle/dv#63516 is merged, a build-scan-summary release must be published before this PR can be finalized. The `buildScanSummaryVersion` bump to that published version is a follow-up commit here — intentionally not included yet, since no definite version exists.

## Verified locally
Against a locally published build-scan-summary (`1.0.8-2026.1.0` in mavenLocal): distribution assembles, the bundled tool accepts `--require-build-scans` (and no longer exposes `--max-total-wait-time`), and BVS emits the flag only when `--fail-if-not-fully-cacheable` is set.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/scripts/lib/build-scan-online.sh
+++ b/components/scripts/lib/build-scan-online.sh
@@ -102,8 +102,11 @@ fetch_build_scan_data() {
     args+=("--brief-logging")
   fi
 
+  # When failing the build if it is not fully cacheable, the Build Scan data
+  # must be available, so instruct the summary tool to wait longer for it to
+  # become available. A user-set max.wait.time in network.settings always wins.
   if [[ "${fail_if_not_fully_cacheable}" == "on" ]]; then
-    args+=("--max-total-wait-time" "120")
+    args+=("--require-build-scans")
   fi
 
   if [[ -n "${run_id}" ]]; then
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,2 +1,1 @@
-- [NEW] TBD
 - [NEW] Add a `max.wait.time` setting to `network.settings` to configure how long to wait for Develocity Build Scan data to become available
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,2 +1,2 @@
 - [NEW] TBD
-- [FIX] Wait longer for Build Scan data to become available when `--fail-if-not-fully-cacheable` is set, and allow the wait time to be configured via `max.wait.time` in `network.settings`
+- [NEW] Add a `max.wait.time` setting to `network.settings` to configure how long to wait for Develocity Build Scan data to become available
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,2 @@
 - [NEW] TBD
+- [FIX] Wait longer for Build Scan data to become available when `--fail-if-not-fully-cacheable` is set, and allow the wait time to be configured via `max.wait.time` in `network.settings`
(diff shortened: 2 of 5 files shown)


components/scripts/lib/build-scan-online.sh
```diff
@@ -102,8 +102,11 @@ fetch_build_scan_data() {
     args+=("--brief-logging")
   fi
 
+  # When failing the build if it is not fully cacheable, the Build Scan data
+  # must be available, so instruct the summary tool to wait longer for it to
+  # become available. A user-set max.wait.time in network.settings always wins.
   if [[ "${fail_if_not_fully_cacheable}" == "on" ]]; then
-    args+=("--max-total-wait-time" "120")
+    args+=("--require-build-scans")
```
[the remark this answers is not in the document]
ribafish: Known and intentional — see the "Dependencies / blockers" section in the PR description. This depends on the build-scan-summary change in gradle/dv#63516; `buildScanSummaryVersion` will be bumped to the release containing `--require-build-scans` as a follow-up commit once that PR is merged and a build-scan-summary version is published. Not resolving until then.

jthurne (approved):

release/changes.md
erichaagdev, line 1: I suggest changing this from "fix" to "new". A new option is now available to configure the wait time for Develocity Build Scan data to become available.

I actually don't think it's accurate as written. Before these changes, we waited 120s when `--fail-if-not-fully-cacheable` was used, and we still wait for 120s after these changes. I just wouldn't mention `--fail-if-not-fully-cacheable` at all.

erichaagdev (changes requested):

release/changes.md
ribafish: Good point — reworded to a [NEW] entry describing the configurable `max.wait.time` setting, and dropped the `--fail-if-not-fully-cacheable` mention since the wait stays 120s there. Fixed in f6ff2a8c.

release/changes.md
```diff
@@ -1 +1,2 @@
 - [NEW] TBD
```
erichaagdev: Now that there is an entry we need to drop this.


```suggestion
```

erichaagdev (changes requested):

